PR #1, opened 2019-07-17
Adding tools for development

maelk:
Adding:
- repo management scripts : init and update
- test script for baremetal provider
- Dockerfile and script to run a development container

### DIFF
--- a/container/Dockerfile
+++ b/container/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:latest
+FROM ubuntu:18.04
 
 RUN apt-get update && \
     apt-get install -y software-properties-common
@@ -11,7 +11,9 @@ RUN add-apt-repository -y ppa:longsleep/golang-backports && \
     curl \
     gnupg-agent \
     golang-go \
-    go-dep
+    golint
+RUN curl -LO https://raw.githubusercontent.com/golang/dep/master/install.sh && \
+    mkdir -p /root/go/bin && chmod +x install.sh && ./install.sh && rm install.sh
 RUN curl -OJL https://github.com/operator-framework/operator-sdk/releases/download/v0.9.0/operator-sdk-v0.9.0-x86_64-linux-gnu && \
     chmod +x operator-sdk-v0.9.0-x86_64-linux-gnu && cp operator-sdk-v0.9.0-x86_64-linux-gnu /usr/local/bin/operator-sdk && rm operator-sdk-v0.9.0-x86_64-linux-gnu
 RUN curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add - && \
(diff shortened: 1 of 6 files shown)


container/run-workspace.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -ue
+
+SCRIPTPATH=$( cd $(dirname $0) >/dev/null 2>&1 ; pwd -P )
```
kanwar-saad: This causes problems of relative paths. you need this to get the absolute path.

"$(dirname "$(readlink -f "$0")")"

tools/init-repo.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+set -ue
+
+SCRIPTPATH=$( cd $(dirname $0) >/dev/null 2>&1 ; pwd -P )
```
kanwar-saad: same here

tools/init-repo.sh
```diff
@@ -0,0 +1,29 @@
...
+  git remote -v
+  # Update "master" on Nordix
+  git fetch upstream
+  git rebase upstream/master
+  git push
+  cd ..
+}
+
+setup_repo cluster-api kubernetes-sigs
+for repo in metal3-dev-env cluster-api-provider-baremetal baremetal-operator
```
kanwar-saad: this is fine for now .. but this script can be modified slightly to make it generic and work for other repos as well.

container/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+    mv ./kubectl /usr/local/bin/kubectl
+
+ENV GOPATH=/root/go
+ENV PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/root/go/bin
+
+RUN curl -LO https://raw.githubusercontent.com/Nordix/cluster-api-provider-baremetal/master/tools/install_kustomize.sh && \
+    chmod +x ./install_kustomize.sh && ./install_kustomize.sh && rm install_kustomize.sh && \
+    curl -LO https://raw.githubusercontent.com/Nordix/cluster-api-provider-baremetal/master/tools/install_kubebuilder.sh && \
+    chmod +x ./install_kubebuilder.sh && ./install_kubebuilder.sh && rm install_kubebuilder.sh && \
+    mv kubebuilder /usr/local
```
namnx228: The `install_kustomize.sh` only installs Kustomize version `1.0.11` but the latest is `3.0.3`. Should we use the newer ?

container/Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM ubuntu:latest
```
kanwar-saad: this usually breaks in CI when there is a new version of ubuntu released with some different package name. locking it to a particular version would be better. like 18.04 tag is updated frequently with security updates and bug fixes so maybe thats fine.

container/Dockerfile
maelk: maybe we could do it in the baremetal-operator kustomize install script. I would not change it here

PR #3, opened 2019-07-24
Reorganize container go tree

maelk:
Move repos around to be correctly located in the go path, install Bazel and change the gitignore to ignore the other repositories and go path

### DIFF
--- a/tools/init-repo.sh
+++ b/tools/init-repo.sh
@@ -20,10 +20,16 @@ setup_repo () {
   cd ..
 }
 
-setup_repo cluster-api kubernetes-sigs
-for repo in metal3-dev-env cluster-api-provider-baremetal baremetal-operator
-do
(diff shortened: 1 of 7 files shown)


tools/init-repo.sh
```diff
@@ -20,10 +20,16 @@ setup_repo () {
   cd ..
 }
 
-setup_repo cluster-api kubernetes-sigs
-for repo in metal3-dev-env cluster-api-provider-baremetal baremetal-operator
-do
-    setup_repo ${repo} metal3-io
-done
+setup_go_repo () {
+  mkdir -p ${3}
```
kanwar-saad: can you quote all variables to avoid globbing. you will save a lot of linter errors laters.
${3} -> "${3}"

maelk: I have another commit coming up for that already ;)

PR #7, opened 2019-07-25
Add a linting container

maelk:


container/run-lint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -ue
+
+SCRIPTPATH="$(dirname "$(readlink -f "${0}")")"
+
+docker build "${SCRIPTPATH}/lint" -t lint
+docker run -v "${SCRIPTPATH}"/../:/data --rm -it --name lint lint
```
kanwar-saad: why -it?

container/lint/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+# (software-properties-common for add-apt-repository)
+RUN apt-get update && \
+    apt-get install -y \
+    git \
+    curl \
+    apt-transport-https \
+    ca-certificates \
+    software-properties-common
+
+#Add repos and install Go, Docker and Bazel
```
kanwar-saad: comment is wrong

container/lint/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+    apt-get update && apt-get install -y \
+    golang-go \
+    golint \
+    shellcheck
+
+ENV GOPATH=/data/go
+ENV PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/data/go/bin
+
+
+#Install Golang dep, operator-sdk, Kubectl, Kustomize and kubebuilder
```
kanwar-saad: same

container/run-lint.sh
kanwar-saad: i guess it should trigger a script which will just run all kind of linters on given folder. like go linters on all go files and python linter on .py files and same for .sh

kanwar-saad:
test this please

kanwar-saad:
retest

container/run-lint.sh
maelk: About triggering a script, that's rather a use of the container than  the container itself. the run-lint.sh script is a convenience script to run it locally with a bash inside. If you want to automate the linting, then another script should be made IMO.

maelk:
@mkth1 is implementing it using separated containers

PR #13, opened 2019-07-29
Add markdown lint

### LINKED ISSUE #12: Add lint for markdown
Add lint for markdown files and fix the markdown errors accordingly

mkth1:
Add support to lint markdown files.
Fix markdown errors in md files.
closes #12

### DIFF
--- a/resources/docker/linter/Dockerfile
+++ b/resources/docker/linter/Dockerfile
@@ -1,0 +1,8 @@
+FROM alpine:3.10
+
+RUN apk update && \
+    apk add --no-cache ruby ruby-json && \
+    gem install --no-rdoc --no-ri mdl && \
+    mkdir /data
+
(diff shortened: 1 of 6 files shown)


resources/docker/linter/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM alpine:3.10
+
+RUN apk update && \
+    apk add --no-cache ruby ruby-json && \
+    gem install --no-rdoc --no-ri mdl && \
+    mkdir /data
+
+WORKDIR /data
```
maelk: Maybe add a CMD to lint the workdir by default ?

mkth1: There is no need for that, since we are calling the command from make target with custom config.

PR #16, opened 2019-07-30
Add shell linter

### LINKED ISSUE #15: Add shell script lint
Add lint for shell scripts and fix the lint errors accordingly

mkth1:
closes #15

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,8 @@ PROJECT := airship
 APP     := tools
 NAME    := ${PROJECT}-${APP}
 
+lint_folder ?= $(CURDIR)
+
 .DEFAULT_HELP := help
 .PHONY: help
 help:
@@ -16,15 +18,19 @@ setup-repos: ## Setup nordix repos
 
 .PHONY: update-repos
 update-repos: ## Update nordix repos
-	@(CURDIR)/scripts/update-nordix-repos-master.sh
+	$(CURDIR)/scripts/update-nordix-repos-master.sh
 
 .PHONY: workspace
 workspace: ## Create and execute dev workspace for nordix repos
-	@(CURDIR)/container/run-workspace.sh
+	$(CURDIR)/container/run-workspace.sh
 
 .PHONY: lint-md
-lint-md: ## Lint markdown
-	docker run --rm -v "${CURDIR}:/data" --name mkdlint ${NAME}-md-lint mdl -s configs/linter/.mdstylerc.rb .
+lint-md: ## Lint markdown (ex: make lint-md or make lint-md lint_folder=abspath)
+	docker run --rm \
+		-v "${CURDIR}:/data" \
+		-v "${lint_folder}:/lintdata" \
+		${NAME}-md-lint \
+		mdl -s configs/linter/.mdstylerc.rb "/lintdata";
 
 .PHONY: build-lint-md
 build-lint-md: ## Build Docker Image for markdown lint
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -2,6 +2,8 @@ PROJECT := airship
 APP     := tools
 NAME    := ${PROJECT}-${APP}
 
+lint_folder ?= *
```
kanwar-saad: This will cause a problem when make is not run from the root directory.
you can use the following line to get absolute path to avoid this problem.
ROOT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))

Makefile
```diff
@@ -22,10 +24,26 @@ update-repos: ## Update nordix repos
 workspace: ## Create and execute dev workspace for nordix repos
 	@(CURDIR)/container/run-workspace.sh
 
+MD_FILES=$(shell find $(lint_folder) -iname '*.md')
```
kanwar-saad: -type f

Makefile
```diff
@@ -22,10 +24,26 @@ update-repos: ## Update nordix repos
...
 .PHONY: lint-md
-lint-md: ## Lint markdown
-	docker run --rm -v "${CURDIR}:/data" --name mkdlint ${NAME}-md-lint mdl -s configs/linter/.mdstylerc.rb .
+lint-md: ## Lint markdown (ex: make lint-md or make lint-md lint_folder=wow)
+	@if [ -z "$(MD_FILES)" ] ; then \
```
kanwar-saad: you can just do

[ -n "${MD_FILES}" ] && docker run ....

Makefile
kanwar-saad: $(ROOT_DIR) is enough instead of lint_folder

Makefile
```diff
@@ -22,10 +24,26 @@ update-repos: ## Update nordix repos
...
+		continue; \
+	else \
+		docker run --rm -v "${CURDIR}:/data" --name mkdlint ${NAME}-md-lint mdl -s configs/linter/.mdstylerc.rb $(MD_FILES); \
```
kanwar-saad: ROOT_DIR instead of CURDIR

Makefile
```diff
@@ -22,10 +24,26 @@ update-repos: ## Update nordix repos
...
+
+SHELLCHECK_VERSION="v0.7.0"
+SHELLCHECK_IMAGE="koalaman/shellcheck:${SHELLCHECK_VERSION}"
+SHELL_FILES=$(shell find $(lint_folder) -iname '*.sh')
+.PHONY: lint-shell
+lint-shell: ## Lint shell scripts (ex: make lint-shell or make lint-shell lint_folder=scripts)
+	@if [ -z "$(SHELL_FILES)" ] ; then \
+		continue; \
+	else \
+		docker run --rm -v "${CURDIR}:/mnt" --name shellcheck ${SHELLCHECK_IMAGE} ${SHELL_FILES}; \
```
kanwar-saad: --name with would be problematic for parallel builds in CI. i suggest to remove --name completely or if you want to give meaningful names then generate a random string and prefix it with "shellcheck" or whatever linter name.

Same comment ... for other places.

Makefile
```diff
@@ -22,10 +24,26 @@ update-repos: ## Update nordix repos
...
+		docker run --rm -v "${CURDIR}:/data" --name mkdlint ${NAME}-md-lint mdl -s configs/linter/.mdstylerc.rb $(MD_FILES); \
+	fi
 
 .PHONY: build-lint-md
 build-lint-md: ## Build Docker Image for markdown lint
 	docker build -t ${NAME}-md-lint -f resources/docker/linter/Dockerfile .
+
+SHELLCHECK_VERSION="v0.7.0"
+SHELLCHECK_IMAGE="koalaman/shellcheck:${SHELLCHECK_VERSION}"
+SHELL_FILES=$(shell find $(lint_folder) -iname '*.sh')
```
kanwar-saad: ROOT_DIR

Makefile
```diff
@@ -22,10 +24,26 @@ update-repos: ## Update nordix repos
...
+		continue; \
+	else \
+		docker run --rm -v "${CURDIR}:/data" --name mkdlint ${NAME}-md-lint mdl -s configs/linter/.mdstylerc.rb $(MD_FILES); \
+	fi
 
 .PHONY: build-lint-md
 build-lint-md: ## Build Docker Image for markdown lint
 	docker build -t ${NAME}-md-lint -f resources/docker/linter/Dockerfile .
+
+SHELLCHECK_VERSION="v0.7.0"
```
kanwar-saad: shouldn't these steps be part of target instead of executing them even if target was different.

Makefile
```diff
@@ -22,10 +24,26 @@ update-repos: ## Update nordix repos
...
+SHELLCHECK_IMAGE="koalaman/shellcheck:${SHELLCHECK_VERSION}"
+SHELL_FILES=$(shell find $(lint_folder) -iname '*.sh')
+.PHONY: lint-shell
+lint-shell: ## Lint shell scripts (ex: make lint-shell or make lint-shell lint_folder=scripts)
+	@if [ -z "$(SHELL_FILES)" ] ; then \
```
kanwar-saad: same here

PR #24, opened 2019-08-02
Add a Dockerfile for image building container

maelk:
Contains:
- packer
- python 3.7
- Openstack client
- jq


maelk:
Added Ansible, make commands and pushed a version to harbor

PR #26, opened 2019-08-05
ci: jumphost lifecycle scripts

kanwar-saad:
Lifecycle scripts for Dev jumphost.


ci/scripts/jumphost/add_jumphost_user.sh
```diff
@@ -0,0 +1,52 @@
+#! /usr/bin/env bash
+
```
maelk: If this file is for creating proxy users and not full fledged users, maybe the name could be adapted ?

PR #27, opened 2019-08-06
ci: Add/Del jumphost SSH keys

kanwar-saad:
Scripts for adding and removing user ssh keys
from artifactory.

- added usage comments in different scripts


maelk:
Maybe creating a directory to store the scripts executed remotely would help to make the structure clearer

PR #37, opened 2019-08-12
ci: Add Makefile targets for pushing images

Jaakko-Os:


Makefile
```diff
@@ -4,6 +4,8 @@ NAME    := ${PROJECT}-${APP}
 
 lint_folder ?= $(CURDIR)
 
+image_registry := "registry.nordix.org/airship/image-builder"
```
maelk: maybe split this like registry_url and the image_name (then specific to the task)

Makefile
```diff
@@ -59,9 +61,9 @@ build-image-builder: ## Build Docker Image for qcow2 image building
 		-f resources/docker/builder/Dockerfile resources/docker/builder/
 
 .PHONY: push-image-builder
-push-image-builder: ## Build Docker Image for qcow2 image building
-	docker tag image-builder registry.nordix.org/airship/image-builder
-	docker push registry.nordix.org/airship/image-builder
+push-image-builder: ## Push Docker Image for qcow2 image building to nordix registry
+	docker tag image-builder ${image_registry}
+	docker push ${image_registry}
```
maelk: keep the image-builder part specific to this task, the registry URL can be parameterized

Makefile
Jaakko-Os: done

Makefile
Jaakko-Os: done

Makefile
```diff
@@ -59,9 +61,9 @@ build-image-builder: ## Build Docker Image for qcow2 image building
 		-f resources/docker/builder/Dockerfile resources/docker/builder/
 
 .PHONY: push-image-builder
-push-image-builder: ## Build Docker Image for qcow2 image building
-	docker tag image-builder registry.nordix.org/airship/image-builder
-	docker push registry.nordix.org/airship/image-builder
+push-image-builder: ## Push Docker Image for qcow2 image building to nordix registry
+	docker tag image-builder ${image_registry}/airship/image-builder
```
kanwar-saad: no version tag like latest

kanwar-saad: docker pull before docker run

kanwar-saad: use full container repo path in run

Jaakko-Os: -tag set via parameter
-full container repo path in use in docker run. Docker run checks if image not available locally, it is downloaded from registry -> separate pull task not needed

Makefile
```diff
@@ -4,6 +4,9 @@ NAME    := ${PROJECT}-${APP}
 
 lint_folder ?= $(CURDIR)
 
+image_registry := registry.nordix.org
```
kanwar-saad: can you change it to container_registry

Makefile
```diff
@@ -4,6 +4,9 @@ NAME    := ${PROJECT}-${APP}
...
+image_version  := v1.0
```
kanwar-saad: image_version is very vague. Since there might be more images coming so it would be nice to tell which image version is this. like md_container_version and you can define them near the targets where it is used so that its easier to group.

Jaakko-Os:
> please do not use the same version for different images

image name is unique so same version does not conflict.
what is the preferred way to tag the images, i.e the version in particular?

Makefile
Jaakko-Os: done

Makefile
Jaakko-Os: missed this. I hope image... is acceptable

nordixinfra:

kanwar-saad:
you need to sign the commit and rebase

Jaakko-Os:
> you need to sign the commit and rebase

done

PR #39, opened 2019-08-13
Fix absolute paths for linter repos

mikkosest:
Absolute paths needed for linter make variables


nordixinfra:

PR #41, opened 2019-08-14
Modify CI setup

maelk:
Reduce size of the jumphost
disable unattended upgrades to fix apt issue
fix variable name issue in integration test scripts


nordixinfra:

PR #44, opened 2019-08-15
Add the hardening script

jan-est:


nordixinfra:

PR #49, opened 2019-08-19
Remove all the unicode characters and add root login in metal3 image

fmuyassarov:


nordixinfra:

PR #51, opened 2019-08-20
 Add condition in image generation script to check if CI keypair doesn't already exists 

fmuyassarov:


nordixinfra:

PR #53, opened 2019-08-20
 Delete airshipci-key if it exists while image generation.

fmuyassarov:


nordixinfra:

PR #59, opened 2019-08-29
ci: gotest container spec for unit test

mikkosest:
Container for unit test

### DIFF
--- a/resources/docker/gotest/Dockerfile
+++ b/resources/docker/gotest/Dockerfile
@@ -1,0 +1,81 @@
+FROM alpine:3.10
+
+RUN apk add --no-cache \
+		ca-certificates
+
+RUN [ ! -e /etc/nsswitch.conf ] && echo 'hosts: files dns' > /etc/nsswitch.conf
+
+ENV GOLANG_VERSION 1.13rc2
+
+RUN set -eux; \
+	apk add --no-cache --virtual .build-deps \
+		bash \
+		gcc \
+		musl-dev \
+		openssl \
+		go \
+	; \
+	export \
+		GOROOT_BOOTSTRAP="$(go env GOROOT)" \
+		GOOS="$(go env GOOS)" \
+		GOARCH="$(go env GOARCH)" \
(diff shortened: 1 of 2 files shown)


resources/docker/gotest/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM golang:1.11-alpine3.7
+
+RUN apk add --update \
+    git \
+    glide \
+    make \
+    sudo \
```
maelk: why sudo ? you have not changed the user (root) so you should not have permission issues. Which script is using it ?

resources/docker/gotest/Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+		/usr/local/go/pkg/obj \
+	; \
+	apk del .build-deps; \
+	\
+	export PATH="/usr/local/go/bin:$PATH"; \
+	go version
+
+ENV GOPATH /go
+
+RUN apk add --update \
```
maelk: why separate the apk steps ?

resources/docker/gotest/Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+    mv /go/bin/dep /usr/local/bin && \
+    curl -LO https://raw.githubusercontent.com/Nordix/cluster-api-provider-baremetal/master/tools/install_kubebuilder.sh && \
+    chmod +x ./install_kubebuilder.sh \
+    && ./install_kubebuilder.sh \
+    && rm ./install_kubebuilder.sh && \ 
+    mv /kubebuilder /usr/local && \
+    cp /usr/local/kubebuilder/bin/kubebuilder /usr/local/bin
+
+COPY run-test.sh /usr/bin/testentrypoint
+ENTRYPOINT ["/usr/bin/testentrypoint"]
```
maelk: maybe CMD rather than ENTRYPOINT ?

mikkosest: Done

PR #65, opened 2019-09-09
Set Go 1.11 modules.

fmuyassarov:


nordixinfra:

resources/docker/workspace/Dockerfile
```diff
@@ -29,7 +29,7 @@ RUN add-apt-repository -y ppa:longsleep/golang-backports && \
 
 ENV GOPATH=/data/go
 ENV PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/data/go/bin
-
+ENV GO111MODULE=on
```
maelk: It might be better to set it where needed, in front of the command line that requires it, rather than globally

PR #71, opened 2019-09-19
Adding openstack image building pipeline

maelk:


ci/jobs/openstack_image_building.pipeline
```diff
@@ -0,0 +1,142 @@
...
+ci_git_credential_id = "nordixinfra-github-token"
+ci_git_branch = "master"
+
+
+pipeline {
+  agent { label 'airship-static-workers' }
+  environment {
+    AIRSHIP_CI_USER="airshipci"
+    RT_URL="https://artifactory.nordix.org/artifactory"
+    OS_AUTH_URL="https://kna1.citycloud.com:5000"
```
kanwar-saad: These OS_* credentials can be taken from credential.

maelk: I don't get what you mean. From the credentials, we get OS_USERNAME and OS_PASSWORD, but we do not get the rest of the OS_* variable, right ?

PR #82, opened 2019-10-04
fix for updating Nordix repositories

Jaakko-Os:
'git remote add' command failed because a whole list was given as an argument and not an item from the list as was ment.


nordixinfra:

PR #86, opened 2019-10-08
ci: add passphrase handling to update Nordix repos pipeline

Jaakko-Os:


nordixinfra:

Jaakko-Os:
ssh-agent will do the trick and handle key passphrase correctly

PR #87, opened 2019-10-08
fix update Nordix repo script when running locally

Jaakko-Os:
set default value for WORKSPACE


nordixinfra:

PR #88, opened 2019-10-11
Kill ci job if integration test faces an error

fmuyassarov:
This patch will force to kill ci job if there is an error&timeout during integration test.


nordixinfra:

PR #93, opened 2019-10-16
Update openstack cli in CI

mkth1:
With os cli update, no need to remove package in metal3 image


nordixinfra:

PR #102, opened 2019-10-30
Add timestamp to catch timeout error on metal3-dev-env integration test

fmuyassarov:
This patch tracks time of a single pipeline stage in order to kill the integration job if timeout happens during the test.

### DIFF
--- a/ci/jobs/capi_bm_integration_tests.pipeline
+++ b/ci/jobs/capi_bm_integration_tests.pipeline
@@ -2,6 +2,7 @@ import java.text.SimpleDateFormat
 
 ci_git_url = "https://github.com/Nordix/airship-dev-tools.git"
 ci_git_credential_id = "nordix-airship-ci-github-prod-token"
+def TIMEOUT = 3600
 
 script {
   if ("${PROJECT_REPO_ORG}" == "Nordix" && "${PROJECT_REPO_NAME}" == "airship-dev-tools") {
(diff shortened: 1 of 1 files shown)


ci/jobs/capi_bm_integration_tests.pipeline
```diff
@@ -84,6 +92,13 @@ pipeline {
   }
   post {
     cleanup {
+      script {
+        CURRENT_BUILD_TIME = System.currentTimeMillis()
+        if ((((CURRENT_BUILD_TIME - CURRENT_START_TIME)/1000) - 3600) > 0) {
```
maelk: Can you parameterize the timeout duration like TIMEOUT*60 and use TIMEOUT variable in the stage ?

fmuyassarov: done

ci/jobs/capi_bm_integration_tests.pipeline
```diff
@@ -37,6 +38,7 @@ pipeline {
     BUILD_URL = "${env.BUILD_URL}"
     PR_ID = "${ghprbPullId}"
     DISTRIBUTION = "${DISTRIBUTION}"
+    TIMEOUT = 3600
```
maelk: Does this need to be defined twice ?

fmuyassarov: No. Removed

PR #114, opened 2019-11-28
Removed instruction that is not longer applicable

Xenwar:


Xenwar:
The is no longer applicable.

PR #116, opened 2019-12-05
Add v1alpha2 integration test pipeline

fmuyassarov:
This PR adds pipeline for v1alpha2 integration test on Metal3-dev-env

### DIFF
--- a/ci/jobs/capi_bm_integration_tests.pipeline
+++ b/ci/jobs/capi_bm_integration_tests.pipeline
@@ -38,6 +38,7 @@ pipeline {
     BUILD_URL = "${env.BUILD_URL}"
     PR_ID = "${ghprbPullId}"
     DISTRIBUTION = "${DISTRIBUTION}"
+    V1ALPHA2_SWITCH = "${V1ALPHA2_SWITCH}"
   }
   stages {
     stage('SCM') {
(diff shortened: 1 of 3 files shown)


ci/jobs/capi_bm_v1a2_integration_tests.pipeline
```diff
@@ -0,0 +1,111 @@
...
+    TEST_EXECUTER_VM_NAME = "${VM_NAME}"
+    BUILD_URL = "${env.BUILD_URL}"
+    PR_ID = "${ghprbPullId}"
+    DISTRIBUTION = "${DISTRIBUTION}"
+    V1ALPHA2_SWITCH=true
+  }
+  stages {
+    stage('SCM') {
+      options {
+        timeout(time: 5, unit: 'MINUTES')
```
Xenwar: define the value 5 as constant, much like the TIMEOUT

ci/jobs/capi_bm_v1a2_integration_tests.pipeline
```diff
@@ -0,0 +1,111 @@
...
+    OS_REGION_NAME="Kna1"
+    OS_PROJECT_NAME="Default Project 37137"
+    OS_TENANT_NAME="Default Project 37137"
+    OS_AUTH_VERSION=3
+    OS_IDENTITY_API_VERSION=3
+    TEST_EXECUTER_VM_NAME = "${VM_NAME}"
+    BUILD_URL = "${env.BUILD_URL}"
+    PR_ID = "${ghprbPullId}"
+    DISTRIBUTION = "${DISTRIBUTION}"
+    V1ALPHA2_SWITCH=true
```
maelk: Do we need a whole new pipeline if this is the only difference ? This could be also achieved by parameterizing the jenkins job, it would avoid duplication

fmuyassarov: yes, that can be done

PR #121, opened 2019-12-22
Add v1a2 default environment variables.

fmuyassarov:
Export memory & distribution default variables for v1a2 integrations test.


ci/scripts/tests/files/run_integration_tests.sh
```diff
@@ -11,6 +11,8 @@ REPO_NAME="${2:-metal3-dev-env}"
 REPO_BRANCH="${3:-master}"
 UPDATED_REPO="${4:-https://github.com/${REPO_ORG}/${REPO_NAME}.git}"
 export CAPI_VERSION="${5:-v1alpha2}"
+export IMAGE_OS="${6:Ubuntu}"
```
maelk: to make it default to something, the syntax is ":-". You have to change four of those

fmuyassarov: done

PR #152, opened 2020-03-08
Clone from source branch for metal3-io.github.io

fmuyassarov:
Clone from `source ` branch of metal3-io.github.io

### DIFF
--- a/scripts/update-nordix-repos-master.sh
+++ b/scripts/update-nordix-repos-master.sh
@@ -72,12 +72,12 @@ do
   echo "Updating master branch in ${repo}"
   pushd "${repo}"
   BRANCH=$(git rev-parse --abbrev-ref HEAD)
-  git checkout "${UPDATE_BRANCH}"
+  git checkout "${BRANCH}"
   # origin points to upstream repos
   git fetch origin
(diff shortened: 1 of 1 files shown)


fmuyassarov:
> It is not sufficient to change the branch for upstream, you also need to change it for Nordix, or you will push the changes from source onto master, and probably get a conflict

Ah good point!

scripts/update-nordix-repos-master.sh
```diff
@@ -72,12 +72,12 @@ do
   echo "Updating master branch in ${repo}"
   pushd "${repo}"
   BRANCH=$(git rev-parse --abbrev-ref HEAD)
-  git checkout "${UPDATE_BRANCH}"
+  git checkout "${BRANCH}"
```
maelk: this needs to be set to source in the metal3 website case. It defaults to master

fmuyassarov: If I didn't miss anything, both Nordix and upstream defaults to source branch in case of metal3 website

fmuyassarov:
> Ok, sorry, reviewed too fast

No worries.

PR #165, opened 2020-05-26
Upgrade boot disk image of the cluster

Xenwar:
Performs upgrade of boot disk image of both controlplane and worker nodes.


Jaakko-Os:
rename file 'upgrade.clusters.h' to 
'upgrade.clusters.sh'

PR #184, opened 2020-06-10
Add variable, small bug fix in common.sh and two other test cases.

smoshiur1237:
small bug fix and added "sudo" command in Upgraded files...


smoshiur1237:
/cc @Jaakko-Os   
please review and marge the changes...

PR #193, opened 2020-06-15
Upgrade test failure handling improved

Jaakko-Os:
If a test script fails due to timeout,
deprovision cluster and continue to the next test.


Xenwar:
I think it handles only timeout reasons. I could be wrong, but anywhere there is `exit #` will kill the test run.
Handle those or change the description accordingly.

PR #197, opened 2020-06-15
Upgrade CAPI version using clusterctl.

Xenwar:


upgrade_tests/controlplane_components_upgrade/upgrade_CAPI_with_clusterctl.sh
```diff
@@ -0,0 +1,84 @@
...
+
+
+# do upgrade
+clusterctl upgrade plan | grep "upgrade apply" | xargs | xargs clusterctl 
+
+# Verify upgrade
+upgraded_controllers_count=$(kubectl api-resources | grep 2020 | wc -l)
+upgraded_bootstrap_crd_count=$(kubectl get crds kubeadmconfigs.bootstrap.cluster.x-k8s.io -o json | jq '.spec.names.singular' | wc -l)
+
+if [ $upgraded_controllers_count != 2 ]
```
Jaakko-Os, line 70: use '-ne' instead of '!='

upgrade_tests/controlplane_components_upgrade/upgrade_CAPI_with_clusterctl.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
+
+set -x
+
+start_logging "${0}"
```
Jaakko-Os, line 5: Since upgrade.sh is calling the test script with 'source scriptname.sh', the script  filename is not discoverable from '${0}' anymore. This worked when calling the script with syntax './scriptname.sh'.
Thus the logfile name shall be given as an argument 'source scriptname.sh testcaselogfilename' and change the function call to 'start_logging "${1}"'

upgrade_tests/controlplane_components_upgrade/upgrade_CAPI_with_clusterctl.sh
Xenwar: done

upgrade_tests/controlplane_components_upgrade/upgrade_CAPI_with_clusterctl.sh
Xenwar: done

upgrade_tests/controlplane_components_upgrade/upgrade_CAPI_with_clusterctl.sh
```diff
@@ -0,0 +1,84 @@
...
+sed -i 's/singular: kubeadmconfig/singular: kubeadmconfig2020/' /home/$USER/.cluster-api/overrides/bootstrap-kubeadm/v0.3.2/bootstrap-components.yaml
+sed -i 's/kcp/kcp2020/' /home/$USER/.cluster-api/overrides/control-plane-kubeadm/v0.3.3/control-plane-components.yaml
+
+
+# do upgrade
+clusterctl upgrade plan | grep "upgrade apply" | xargs | xargs clusterctl 
+
+# Verify upgrade
+upgraded_controllers_count=$(kubectl api-resources | grep 2020 | wc -l)
+upgraded_bootstrap_crd_count=$(kubectl get crds kubeadmconfigs.bootstrap.cluster.x-k8s.io -o json | jq '.spec.names.singular' | wc -l)
```
kashifest, line 68: Any specific reason to check only bootstrap only? What is this crd count by the way

Xenwar: It is checking all three.
The first two, the change is made of shortNames
The kubeadmconfig does not such such naming. For it the check is done on 'singular' name of the resources.

PR #204, opened 2020-06-22
[WIP] Upgrade controllers, k8s version, and node images in one shot.

Xenwar:


Xenwar:
This has been move to metal3-dev-env
All except k8s version are move. We have decided to add 'k8s version' and should be included there.

PR #208, opened 2020-06-25
Improve upgrade tests before they are moved to metal3-dev-env

Xenwar:


Xenwar:
Relevant test cases from this PR are moved to metal3-dev-env.
This PR serves as a back for upgrade test cases. Should we decide to add more cases, we take them from here.

Due to that, I am force merging this PR.

PR #216, opened 2020-07-28
Remove password policies

maelk:
since we do not use passwords in CI vms


fmuyassarov:
> What about `COMMON_PASS_PATH`, do we need it?

no, actually both need to be cleaned.

maelk:
done thank you!

PR #217, opened 2020-07-30
Fix centos error with minikube

furkatgofurov7:
First try to fix [Jenkins /test-v1a3-centos-integration fail](https://jenkins.nordix.org/view/Airship/job/airship_metal3io_project_infra_v1a3_integration_test_centos/32/console) in [pull/126](https://github.com/metal3-io/project-infra/pull/126)

### DIFF
--- a/ci/scripts/image_scripts/provision_metal3_image_centos.sh
+++ b/ci/scripts/image_scripts/provision_metal3_image_centos.sh
@@ -45,6 +45,9 @@ if [[ "${DEPLOY_METAL3}" == "true" ]]; then
   git checkout "${M3_DENV_BRANCH}"
   git pull -r || true
   make install_requirements
+  sudo minikube stop
+  sudo minikube delete
+  for veth in $(ifconfig | grep -w "provisioning" | grep -w "baremetal"); do ifconfig $veth delete; done
   popd
 
   rm -rf "${M3_DENV_PATH}"


ci/scripts/image_scripts/provision_metal3_image_centos.sh
```diff
@@ -45,6 +45,9 @@ if [[ "${DEPLOY_METAL3}" == "true" ]]; then
   git checkout "${M3_DENV_BRANCH}"
   git pull -r || true
   make install_requirements
+  sudo minikube stop
+  sudo minikube delete
+  for veth in $(ifconfig | grep "^veth" | cut -d' ' -f1); do ifconfig $veth down; done
```
maelk, line 50: rather delete it maybe ? and you should know exactly what the names of the veth are

furkatgofurov7: @maelk agree, delete should be used. I gave it a try, would that make sense to you?

fmuyassarov: I'm not sure if you can really know that name of the minikube interface since it creates an interface names veth+7 random characters. 
```
15: vetha4a592c@if14: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master docker0 state UP group default 
    link/ether 66:f3:90:6c:c8:cf brd ff:ff:ff:ff:ff:ff link-netnsid 1
    inet6 fe80::64f3:90ff:fe6c:c8cf/64 scope link 
       valid_lft forever preferred_lft forever
```

And I don't think that we need to delete the interface manually since minikube delete should take of it.

ci/scripts/image_scripts/provision_metal3_image_centos.sh
```diff
@@ -45,6 +45,9 @@ if [[ "${DEPLOY_METAL3}" == "true" ]]; then
   git checkout "${M3_DENV_BRANCH}"
   git pull -r || true
   make install_requirements
+  sudo minikube stop
+  sudo minikube delete
+  for veth in $(ifconfig | grep -w "provisioning" | grep -w "baremetal"); do ifconfig $veth delete; done
```
fmuyassarov, line 50: Maybe we don't need to delete it manually? 
@furkatgofurov7 , @maelk ?

PR #225, opened 2020-09-08
Remove test user and revert back

furkatgofurov7:
Removing test user which was added for debugging purposes and reverting everything back to its original state.

### DIFF
--- a/ci/scripts/image_scripts/provision_metal3_image_centos.sh
+++ b/ci/scripts/image_scripts/provision_metal3_image_centos.sh
@@ -51,9 +51,6 @@ if [[ "${DEPLOY_METAL3}" == "true" ]]; then
   rm -rf "${M3_DENV_PATH}"
 fi
 
-sudo useradd testuser1 || true
-echo 'testuser1:newpassword' | sudo chpasswd
-sudo usermod -aG wheel testuser1
 sudo sed -i "0,/.*PermitRootLogin.*/s//PermitRootLogin yes/" /etc/ssh/sshd_config
 
 # Reset cloud-init to run on next boot.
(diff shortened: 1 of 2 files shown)


ci/scripts/image_scripts/provision_metal3_image_centos.sh
```diff
@@ -51,10 +51,10 @@ if [[ "${DEPLOY_METAL3}" == "true" ]]; then
   rm -rf "${M3_DENV_PATH}"
 fi
 
-sudo useradd testuser1 || true
-echo 'testuser1:newpassword' | sudo chpasswd
-sudo usermod -aG wheel testuser1
+# Unlock password for airshipci user
+sudo passwd -u "airshipci" -f
```
maelk, line 55: Can you rather get the current user ? like $USER .  This would avoid hardcoding something that is parameterized somewhere else

furkatgofurov7: Done

ci/scripts/image_scripts/provision_metal3_image_centos.sh
```diff
@@ -51,10 +51,10 @@ if [[ "${DEPLOY_METAL3}" == "true" ]]; then
   rm -rf "${M3_DENV_PATH}"
 fi
 
-sudo useradd testuser1 || true
-echo 'testuser1:newpassword' | sudo chpasswd
-sudo usermod -aG wheel testuser1
+# Unlock password for airshipci user
+sudo passwd -u "${USER}" -f
+
 sudo sed -i "0,/.*PermitRootLogin.*/s//PermitRootLogin yes/" /etc/ssh/sshd_config
```
Xenwar, line 57: I don't think we need this. The key for root is not sent to the VM

furkatgofurov7: This was there originally and this PR does not touch that configuration

maelk (commented):
lgtm

Xenwar (approved):

ci/scripts/image_scripts/provision_metal3_image_centos.sh
maelk: we need root login permission for metal3-dev-env. vbmc and sushy-tools use that to communicate with libvirt

PR #227, opened 2020-09-16
Fix image flavor variable naming

kashifest:
This PR fixes the image flavor variable name mismatch

### DIFF
--- a/ci/images/gen_base_ubuntu_image.sh
+++ b/ci/images/gen_base_ubuntu_image.sh
@@ -58,7 +58,7 @@ packer build \
   -var "network=${NETWORK}" \
   -var "floating_ip_net=${FLOATING_IP_NETWORK}" \
   -var "local_scripts_dir=${SCRIPTS_DIR}" \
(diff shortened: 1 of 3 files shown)


Jaakko-Os:
I don't see any fix here. 'image_flavor' -> 'flavor', why is the change needed?

kashifest:
> I don't see any fix here. 'image_flavor' -> 'flavor', why is the change needed?

That is because you are not looking into the whole file probably, the default value is here, 

https://github.com/Nordix/airship-dev-tools/blob/master/ci/images/image_builder_template.json#L14

Jaakko-Os:
> > I don't see any fix here. 'image_flavor' -> 'flavor', why is the change needed?
> 
> That is because you are not looking into the whole file probably, the default value is here,
> 
> https://github.com/Nordix/airship-dev-tools/blob/master/ci/images/image_builder_template.json#L14

That clarifies it, thanks.

Jaakko-Os (approved):

furkatgofurov7 (commented):
/lgtm

PR #229, opened 2020-09-21
Increase volume size in ubuntu volume generating script

furkatgofurov7:
Increase volume size from 50Gb to 200Gb


maelk:
I am not sure this is a good idea, considering that the bigger the volume is, the longer it takes to copy it. I would actually recommend to size it down to 20 GB. What is the reason for increasing the size ?

furkatgofurov7:
@maelk please check this: https://github.com/metal3-io/project-infra/pull/147#issuecomment-696271286

maelk:
Can we close this now that we have a proper way to resize the image ?

furkatgofurov7:
> Can we close this now that we have a proper way to resize the image ?

Yes sure, we can close this, as no need for increase in size. 

/close

PR #230, opened 2020-09-23
Add script to shrink,resize and upload node image

kashifest:
This PR adds functionality to download a node image from openstack, performs shrink,resize operations on the image and upload node image to artifactory using a script.

### DIFF
--- a/ci/images/gen_ubuntu_node_image.sh
+++ b/ci/images/gen_ubuntu_node_image.sh
@@ -5,9 +5,6 @@ set -eu
 SSH_PRIVATE_KEY_FILE="${1:?}"
 USE_FLOATING_IP="${2:?}"
 
-
-
-
 CI_DIR="$(dirname "$(readlink -f "${0}")")/.."
 IMAGES_DIR="${CI_DIR}/images"
 SCRIPTS_DIR="${CI_DIR}/scripts/image_scripts"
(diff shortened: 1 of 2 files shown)


ci/scripts/artifactory/upload.sh
```diff
@@ -0,0 +1,58 @@
+#! /bin/bash
```
maelk, line 1: this filename does not represent what is in the script. Please add some info about what it does

ci/scripts/artifactory/upload.sh
```diff
@@ -0,0 +1,58 @@
...
+#
+set -x
+IMAGE_NAME="${1:?}"
+
+CI_DIR="$(dirname "$(readlink -f "${0}")")/../.."
+RT_SCRIPTS_DIR="${CI_DIR}/scripts/artifactory"
+export KUBERNETES_VERSION=${KUBERNETES_VERSION:-"v1.18.8"}
+
+#check if necessary package is installed or not
+PKG_OK=$(dpkg-query -W --showformat='${Status}\n' libguestfs-tools|grep "install ok installed")
```
maelk, line 23: do you need to check it ? you could just run the install step, and it will just say that it is already installed, no ?

ci/scripts/artifactory/upload.sh
```diff
@@ -0,0 +1,58 @@
...
+mkdir -p "$WORK_DIR"
+echo "Downloading node Image from openstack"
+openstack image save --file "${WORK_DIR}"/"$IMAGE_NAME".qcow2 "$IMAGE_NAME"
+IMAGE_NAME="$IMAGE_NAME".qcow2
+
+# Resize image  
+pushd "${WORK_DIR}"
+export LIBGUESTFS_DEBUG=1 LIBGUESTFS_TRACE=1
+ls -la
+sudo virt-sparsify "$IMAGE_NAME" converted_"$IMAGE_NAME" --compress
```
maelk, line 39: virt-sparsify can operate in place, then you do not need the following two lines (that could also just be the second one, as mv would overwrite the existing file)

maelk (commented):
looks good! Just some nits inline

ci/scripts/artifactory/upload.sh
kashifest: it can, but it cannot do compress for in place, I have tried that

ci/scripts/artifactory/upload.sh
kashifest: Just thought, its fancy that way :) , I can remove it if you want

ci/scripts/artifactory/upload.sh
kashifest: You mean the filename itself should be changed or should I add more description about what it does? I thought keeping upload in artifactory was kinda self explanatory

ci/scripts/artifactory/upload.sh
kashifest: Done

ci/scripts/artifactory/upload.sh
kashifest: Done

ci/scripts/artifactory/upload.sh
maelk: ah true!

maelk (approved):

PR #232, opened 2020-10-06
Fix condition checking when volume status contains an error substring

furkatgofurov7:
When creating a volume, volume might go into error status and stay there due to the improper condition check. This PR fixes that, by checking if volume status contains the substring error.


furkatgofurov7:
Same fix for condition checking as in [metal3-io/project-infra/pull/160]( https://github.com/metal3-io/project-infra/pull/160#)

maelk (approved):

fmuyassarov (approved):

PR #252, opened 2020-12-22
Remove unnecessary code pieces

kashifest:
1. There is no volume build in normal openstack image building, so
volume size is unnecessary
2. We do not need to hold kubelet,kubeadm,kubectl. That was an
experimental code.


smoshiur1237 (approved):
Looks good to me.
/approve

namnx228:
lgtm

PR #254, opened 2021-01-25
Uplift k8s to 1.20.2

kashifest:


fmuyassarov (approved):

ci/scripts/image_scripts/install_crio_on_centos.sh
```diff
@@ -2,7 +2,7 @@
 
 set -uex
 
-export CRIOCTL_VERSION=${CRICTL_VERSION:-"v1.20.0"}
+export CRICTL_VERSION=${CRICTL_VERSION:-"v1.20.0"}
 OS=${OS:-"CentOS_8"}
 # CRI-O version goes 1:1 with Kubernetes version. Thus,
```
furkatgofurov7, line 7: As per comment, are not we supposed to uplift cri-o version to 1.20.2 as well?

kashifest: crictl version is different than k8s, https://github.com/kubernetes-sigs/cri-tools/releases

furkatgofurov7: Apparently, latest version seems to be [**v1.19.1**](https://github.com/cri-o/cri-o/releases) ? @fmuyassarov

smoshiur1237: v1.20.0 is the latest version of cri-o. I couln't find 1.20.2 version.

furkatgofurov7: Sorry, then it is on line 11, not sure how important it is to make sure 1:1 version compatibility between `KUBERNETES_VERSION` and `VERSION`

kashifest: crioctl should be compatible with minor release, I mean k8s 1.20.x should be compatible with crio 1.20. Since crio doesn't release a new version for every patch version 'x' of k8s, are you @furkatgofurov7 proposing to stick to k8s minor version only and don't update to every patch version?

furkatgofurov7: I was concerned about commment message only stating that `VERSION` of CRI-O should map 1:1 to K8s version which currently maps, but after this change it does not. I was just thinking of it, how important to keep them in sync. Not blocking this patch at all.

kashifest: I understand your concern and I didnt mean you are blocking @furkatgofurov7 :D , @fmuyassarov whats your thought on this?

fmuyassarov: I think, I think 1:1 mapping doesn't necessarily require minor version to match, only major versions should be good enough. One more thing, `CRICTL` is for client, and is separate from `CRI-O` itself. So, mapping is between `CRI-O` & `K8S`.

fmuyassarov: > Apparently, latest version seems to be [**v1.19.1**](https://github.com/cri-o/cri-o/releases) ? @fmuyassarov

Github releases page shows in the order, but not the latest release. You can see here that they have 1.20 release
https://github.com/cri-o/cri-o/releases/tag/v1.20.0

PR #259, opened 2021-02-04
Start the CRI-O before pulling the images

fmuyassarov:
Start the CRI-O before pulling the images.

### DIFF
--- a/ci/scripts/image_scripts/source_cluster_container_images.sh
+++ b/ci/scripts/image_scripts/source_cluster_container_images.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+set -uex
+
+export CONTAINER_RUNTIME="${CONTAINER_RUNTIME}"
 # Container images that we pre-pull into the CI image.
 export IMG_REGISTRY="${IMG_REGISTRY:-"docker.io/registry:latest"}"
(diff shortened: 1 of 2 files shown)


ci/scripts/image_scripts/source_cluster_container_images.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+set -uex
+
+export CONTAINER_RUNTIME="${CONTAINER_RUNTIME}"
```
kashifest, line 5: no default?

fmuyassarov: no, because it will be taken from the value given in the script which is calling this (provision_image.sh)

fmuyassarov: I don't want to put the default here, because this script is called both by Ubuntu and Centos image builder scripts. And they have different runtimes.

fmuyassarov: Any suggestions?

kashifest: its fine I guess since its dependent on the caller

kashifest (approved):

PR #266, opened 2021-02-19
Move the Calico to v3.17.3

fmuyassarov:
Move the Calico to [v3.17.3](https://github.com/projectcalico/calico/releases/tag/v3.17.3)


namnx228 (approved):
LGTM. Do we also need the same change for metal3-dev-env?

PR #267, opened 2021-02-21
Bump the Golang version to 1.16

fmuyassarov:
Bump the Golang version to [1.16](https://blog.golang.org/go1.16). This is needed to make sure that we download the right version of the Golang container images that will be used in BMO/CAPM3/IPAM repos while building the controller images. This PR should be the last one to merged after.


Xenwar (approved):

furkatgofurov7 (commented):
/lgtm

PR #273, opened 2021-04-06
Add cargo dependency to image-builder Dockerfile

macaptain:
This fixes an issue where the Python cryptography package would not
build when running `make build-image-builder. The Rust compiler has been
added as a dependency for building this Python package.

See: https://github.com/pyca/cryptography/issues/5771


namnx228 (approved):
LGTM

PR #274, opened 2021-04-06
Add renovate configuration file into this repo

namnx228:
This PR adds renovate.json configuration file to the repo, so Renovate can work with it.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,0 +1,49 @@
+{
+ "extends": ["config:base"],
+ "baseBranches": ["master"],
+ "bumpVersion": null,
+ "packageRules": [
+   {
+     "depTypeList": ["devDependencies"],
+     "updateTypes": ["minor"],
+     "automerge": false
+   }
+ ],
+ "prHourlyLimit": 5,
+ "prConcurrentLimit": 10,
+ "rangeStrategy": "bump",
+ "renovateFork": true,
+ "includeForks": true,
+ "regexManagers": [
+    {
+      "fileMatch": [".*$"],
+      "matchStrings": ["docker.io/calico/cni:(?<currentValue>v.*?)\"}"],
+      "depNameTemplate": "calico/cni",
+      "datasourceTemplate": "docker"
+    },
+    {
+      "fileMatch": [".*$"],
+      "matchStrings": ["docker.io/calico/pod2daemon-flexvol:(?<currentValue>v.*?)\"}"],
+      "depNameTemplate": "calico/pod2daemon-flexvol",
+      "datasourceTemplate": "docker"
+    },
+    {
+      "fileMatch": [".*$"],
+      "matchStrings": ["docker.io/calico/node:(?<currentValue>v.*?)\"}"],
+      "depNameTemplate": "calico/node",
+      "datasourceTemplate": "docker"
+    },
+    {
+      "fileMatch": [".*$"],
+      "matchStrings": ["docker.io/calico/kube-controllers:(?<currentValue>v.*?)\"}"],
+      "depNameTemplate": "calico/kube-controllers",
+      "datasourceTemplate": "docker"
+    },
+    {
+      "fileMatch": ["^ci/.*$"],
+      "matchStrings": ["CRICTL_VERSION:-\"(?<currentValue>v.*?)\"}"],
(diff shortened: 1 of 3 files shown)


renovate.json
```diff
@@ -0,0 +1,55 @@
...
+      "matchStrings": ["KUBERNETES_VERSION:-\"(?<currentValue>v.*?)\"}"],
+      "depNameTemplate": "kubernetes/kubernetes",
+      "datasourceTemplate": "github-tags"
+    },
+    {
+      "fileMatch": ["Makefile$"],
+      "matchStrings": ["SHELLCHECK_VERSION := \"(?<currentValue>v.*?)\""],
+      "depNameTemplate": "koalaman/shellcheck-alpine",
+      "datasourceTemplate": "docker"
+    },
```
fmuyassarov, line 47: I think we don't prepull this image in [here](https://github.com/Nordix/airship-dev-tools/blob/fae5fafcaf47ebbc9269b14140df2f255d3a79a0/ci/scripts/image_scripts/source_cluster_container_images.sh#L6), because this is not used by integration CI, it is only prow and user who can benefit from this image. We should probably skip it here?

renovate.json
```diff
@@ -0,0 +1,55 @@
...
+      "matchStrings": ["docker.io/calico/kube-controllers:(?<currentValue>v.*?)\"}"],
+      "depNameTemplate": "calico/kube-controllers",
+      "datasourceTemplate": "docker"
+    },
+    {
+      "fileMatch": ["^ci/.*$"],
+      "matchStrings": ["KUBERNETES_VERSION:-\"(?<currentValue>v.*?)\"}"],
+      "depNameTemplate": "kubernetes/kubernetes",
+      "datasourceTemplate": "github-tags"
+    },
```
fmuyassarov, line 41: Would you mind to share where do we use it?

renovate.json
```diff
@@ -0,0 +1,55 @@
...
+      "matchStrings": ["docker.io/calico/pod2daemon-flexvol:(?<currentValue>v.*?)\"}"],
+      "depNameTemplate": "calico/pod2daemon-flexvol",
+      "datasourceTemplate": "docker"
+    },
+    {
+      "fileMatch": [".*$"],
+      "matchStrings": ["docker.io/calico/kube-controllers:(?<currentValue>v.*?)\"}"],
+      "depNameTemplate": "calico/kube-controllers",
+      "datasourceTemplate": "docker"
+    },
```
fmuyassarov, line 35: You are missing docker.io/calico/node [image](https://github.com/Nordix/airship-dev-tools/blob/70db48bf4e4850102637e38bc853256d4b9c1df7/ci/scripts/image_scripts/target_cluster_container_images.sh#L8).

fmuyassarov (changes requested):

renovate.json
namnx228: You can find the places where this item is updated here: https://github.com/namnx228/airship-dev-tools/pull/14/files

fmuyassarov: I see. Who will ensure that once `KUBERNETES_VERSION` is updated here, the same change is made in https://github.com/metal3-io/metal3-dev-env/search?q=KUBERNETES_VERSION ?

namnx228: Good question. I think we also need to use renovate for metal3-dev-env. Otherwise, we may need to update the Kubernetes version in the repo ourshelf, or we should ignore any PR that updates K8s version in this repo.

macaptain (approved):
Let's try Renovate!

renovate.json
furkatgofurov7: @namnx228 we have to set up renovate in m3-dev-env too and maybe in other repos in the future, you are right, and mostly probably IMO, we could use [`Shareable Config Presets`](https://docs.renovatebot.com/config-presets/) feature of renovate to distribute configuration that can be reused across multiple repositories. /cc @macaptain

macaptain: @furkatgofurov7 we could certainly try it. Although I expect the metal3-dev-env renovate config will need to differ at least slightly if we run Renovate from the Nordix repo but the target of the pull request is the upstream repo owned by the metal3 org.

macaptain (approved):

fmuyassarov (approved):
Thank you!

PR #275, opened 2021-04-09
Uplift to k8s 1.21.0

kashifest:


fmuyassarov (commented):
I think this needs to be on hold until 1.21.1 is released?

fmuyassarov:
Is it only for building node images with 1.21? And they will be used by Upgrade tests ?

kashifest:
> I think this needs to be on hold until 1.21.1 is released?

This will create image only for testing. So this can go in IMO. 1.20.4 image is already built. We can keep in hold also.

kashifest:
> Is it only for building node images with 1.21? And they will be used by Upgrade tests ?

That was the idea. For upgrade we dont need the image. I was more interested to test how it works in our case. For that I already did a manual ubuntu image build actually. So we can keep this on hold

macaptain:
I fully expect we will need to deploy a patched `kubeadm` again to workaround the multi-control plane etcd issue (https://github.com/kubernetes/kubernetes/issues/99305).

So the idea is:
1. Branch off kubernetes at 1.21.0 tag
2. Apply this patch: https://github.com/Nordix/kubernetes/commit/4f20e79b4494373dbf5242ff8d1aa34c369776c2
3. Build and push patched kubeadm
4. Add 1.21.0 as an affected version here: https://github.com/Nordix/airship-dev-tools/blob/ecf23f8937376889cff118daf9885742c0f3dc7d/ci/scripts/image_scripts/provision_node_image.sh#L43-L51

macaptain (approved):

PR #292, opened 2021-05-03
Fix bug that all users in the owner list are asked for review

namnx228:


kashifest (approved):

renovate.json
```diff
@@ -46,8 +46,6 @@
       "datasourceTemplate": "github-tags"
     }
   ],
-  "assigneesFromCodeOwners": true,
-  "assigneesSampleSize": 1,
-  "reviewersFromCodeOwners": true,
-  "reviewersSampleSize": 1
+  "assignees": ["kashifest", "fmuyassarov","furkatgofurov7", "Xenwar", "smoshiur1237", "macaptain", "jan-est", "namnx228"],
```
kashifest, line 49: Instead of adding individual github handles, why don't we use  [this](https://github.com/orgs/Nordix/teams/airship) ? In that scenario we dont have to manually maintain the list but whosoever is in github org will be asked to review. Not sure of the sample size would work on it or not though.

namnx228: I think the sample size would not work on it because Renovatebot can only assign to "at"airship group, then Github (not Renovate) will send notifications to all people in the group. But your point makes sense, @kashifest. So if we are okay when Renovate spams notifications to everyone for every PR, the airship group can be used. Otherwise, I think we should keep the current approach.

kashifest: either way is fine with me. Its not a big deal. Its just that current way would require manual maintenance once someone joins in/goes away. I ll approve it.

kashifest (approved):

renovate.json
namnx228: Thanks for your approval

PR #302, opened 2021-05-07
Make /etc/resolv.conf immutable

Xenwar:
Prevents Network Manager from adding nameserver entries in /etc/resolv.conf

### DIFF
--- a/ci/images/userdata.tpl
+++ b/ci/images/userdata.tpl
@@ -9,3 +9,8 @@ users:
 
 runcmd:
   - sed -i "/^127.0.0.1/ s/$/ ${HOSTNAME}/" /etc/hosts
+  - echo -e "nameserver 8.8.8.8" > /etc/resolv.conf
+  # Make /etcd/resolv.conf immutable in order to prevent NetworkManager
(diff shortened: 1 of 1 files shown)


namnx228:
LGTM, just curious: If someone tries to modify the file, but fails to do that, can it cause any errors?

ci/images/userdata.tpl
```diff
@@ -9,3 +9,6 @@ users:
 
 runcmd:
   - sed -i "/^127.0.0.1/ s/$/ ${HOSTNAME}/" /etc/hosts
+  - echo -e "nameserver 127.0.0.53\nnameserver 8.8.8.8" > /etc/resolv.conf
```
macaptain, line 12: Not sure that 127.0.0.53 is required, because systemd-resolved is inactive in the CentOS image. 

Would it also be worth leaving a comment here in the file to explain that the file is immutable by default in order to prevent NetworkManager from overriding DNS with values from DHCP servers?

Xenwar: Thanks. Done

macaptain (approved):

PR #304, opened 2021-05-18
Make /etc/resolve.conf immutable

Xenwar:
Make /etcd/resolv.conf immutable in order to prevent NetworkManager
from overriding DNS entries with values from DHCP servers
This removes network settings of the image building environment from the image.


macaptain (approved):
lgtm

PR #306, opened 2021-05-20
Configure e2e environment

Xenwar:


capo/Makefile
```diff
@@ -21,8 +22,10 @@ openstack:
 		-v /tmp/openstackrc:/tmp/openstackrc \
 		openstacktools/openstack-client \
 		/usr/bin/env BASH_ENV=/tmp/openstackrc /bin/bash -c openstack
+capo-e2e:
+
 deprovision:
-	kubectl delete cluster $(CLUSTER_NAME) --timeout=5m --ignore-not-found=true
+	kubectl delete cluster $(CLUSTER_NAME) --timeout=5m --ignore-not-found=true || echo "No cluster found"
```
macaptain, line 28: We already have `--ignore-not-found=true`, so we don't need the `|| echo`.

capo/Makefile
```diff
@@ -21,8 +22,10 @@ openstack:
 		-v /tmp/openstackrc:/tmp/openstackrc \
 		openstacktools/openstack-client \
 		/usr/bin/env BASH_ENV=/tmp/openstackrc /bin/bash -c openstack
+capo-e2e:
```
macaptain, line 25: Should there be a script associated with this `make` command?

capo/configure.sh
```diff
@@ -53,7 +54,20 @@ echo "Configuring CAPO deployment for project: ${OS_PROJECT_NAME}"
 
 cat << EOF >/tmp/clouds.yaml
 clouds:
-  openstack:
+  "${OS_CLOUD_NAME}":
```
macaptain, line 57: Do we need two separate clouds in clouds.yaml? I don't particularly mind what our cloud is called. If it has to be called capo-e2e for the e2e tests to work, I would be happy to have only one cloud here.

Xenwar: I agree. Will remove capo-e2e clouds. 
or this needs to be static as well  "${OS_CLOUD_NAME}" ?

macaptain: As far as I can tell, the cloud name doesn't have any impact on the final cluster and it's an implementation detail, so this shouldn't be configurable in our dev environment.

macaptain (dismissed):

macaptain (approved):

PR #317, opened 2021-06-08
disable requiretty for ubuntu user

Xenwar:
This PR changes the ssh configuration and sudoers file to bypass issues related to running the e2e tests

### DIFF
--- a/ci/scripts/image_scripts/hardening_base_image.sh
+++ b/ci/scripts/image_scripts/hardening_base_image.sh
@@ -10,10 +10,11 @@ declare -A ssh_values=(
   [MaxAuthTries]=5
   [Ciphers]="aes128-ctr,aes192-ctr,aes256-ctr"
   [ClientAliveInterval]=900
-  [ClientAliveCountMax]=0
+  [ClientAliveCountMax]=15
   [UsePAM]=yes
   [Protocol]=2
 )
+
 
 # Parameters to secure networking /etc/sysctl.conf
 declare -A network_parameters=(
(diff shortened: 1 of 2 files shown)


ci/scripts/image_scripts/hardening_base_image.sh
```diff
@@ -10,11 +10,12 @@ declare -A ssh_values=(
   [MaxAuthTries]=5
   [Ciphers]="aes128-ctr,aes192-ctr,aes256-ctr"
   [ClientAliveInterval]=900
-  [ClientAliveCountMax]=0
+  [ClientAliveCountMax]=15
```
lentzi90, line 13: Why is this needed? If I understand correctly this means the server will wait for `ClientAliveInterval` seconds before checking if the client is alive. If it doesn't get a response after `ClientAliveCountMax` it closes the connection. With `ClientAliveInterval=900` and `ClientAliveCountMax=0` this would be after 900 s = 15 min. With the new values it is 3.5 hours. Do we need this?

Xenwar: No, we do not. ClientAliveInterval was already there and the ClientAliveCountMax is being added to see it solves the following problem.
During E2E tests, an ssh command is run via a tunnel (a -> b -> c). The code run in (a) and executed on (b). The command is executed successfully on C (eg. touch file1.txt). However, the connection terminated immediately. 

It seemed to me that either the client or the server or both are misconfigured in our images/code. So, ClientAliveCountMax is one such property.

Xenwar:
This may not be the right solution.
The following are some alternatives:
Disable the usage of bastion host.
Use the same image as CAPO where the e2e has succeeded. (eg. e2e in the GCP infra.)

PR #318, opened 2021-06-09
Removes target and script that remove old openstack resources

Xenwar:
Removes automated removal of openstack resources.
Users need to remove such resources manually


Xenwar:
This PR is no longer needed. The deletion script is made to be interactive in https://github.com/Nordix/airship-dev-tools/pull/323

PR #319, opened 2021-06-10
Use ubuntu user for running test as there is no need to user cirros 

Xenwar:
> Use ubuntu user for running test as there is no need to user cirros user
> Also adds quotes around a public key

### DIFF
--- a/capo/run-e2e-test.sh
+++ b/capo/run-e2e-test.sh
@@ -3,7 +3,7 @@
 GIT_ROOT=$(git rev-parse --show-toplevel)
 CAPO_REPO=$(realpath "${GIT_ROOT}/../cluster-api-provider-openstack")
 CAPO_E2E_SSH_KEY="${CAPO_REPO}/_artifacts/ssh/cluster-api-provider-openstack-sigs-k8s-io.pub"
-echo "export OPENSTACK_SSH_AUTHORIZED_KEY=$(cat $CAPO_E2E_SSH_KEY)" >> /tmp/e2e_vars_openstack.sh
+echo export OPENSTACK_SSH_AUTHORIZED_KEY='"'$(cat $CAPO_E2E_SSH_KEY)'"' >> /tmp/e2e_vars_openstack.sh
 
 echo "source relevant variables"
 source ./capo_os_vars.rc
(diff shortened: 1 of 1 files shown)


capo/run-e2e-test.sh
```diff
@@ -3,7 +3,7 @@
 GIT_ROOT=$(git rev-parse --show-toplevel)
 CAPO_REPO=$(realpath "${GIT_ROOT}/../cluster-api-provider-openstack")
 CAPO_E2E_SSH_KEY="${CAPO_REPO}/_artifacts/ssh/cluster-api-provider-openstack-sigs-k8s-io.pub"
-echo "export OPENSTACK_SSH_AUTHORIZED_KEY=$(cat $CAPO_E2E_SSH_KEY)" >> /tmp/e2e_vars_openstack.sh
+echo export OPENSTACK_SSH_AUTHORIZED_KEY='"'$(cat $CAPO_E2E_SSH_KEY)'"' >> /tmp/e2e_vars_openstack.sh
```
lentzi90, line 6: I found this a bit hard to understand. Is the idea to make sure that we get quotes around the value of the variable?

```bash
# Before the value was not quoted
export OPENSTACK_SSH_AUTHORIZED_KEY=value here
# Now it is quoted
export OPENSTACK_SSH_AUTHORIZED_KEY="value here"
```
Maybe we could add quotes also for `echo`? I mean it will probably work without but looks a bit strange for me at least.

```suggestion
echo 'export OPENSTACK_SSH_AUTHORIZED_KEY='"'$(cat $CAPO_E2E_SSH_KEY)'"'' >> /tmp/e2e_vars_openstack.sh
```

wgslr: Is the goal that /tmp/e2e_vars_openstack.sh has the literal contents of

Xenwar: Yes. The key pairs are deleted-created for each test run

Xenwar:
@lentzi90 with single quote around the echo, 
`echo 'export OPENSTACK_SSH_AUTHORIZED_KEY='"'$(cat $CAPO_E2E_SSH_KEY)'"''`
Gives
`export OPENSTACK_SSH_AUTHORIZED_KEY='ssh-rsa .....Generated-by-Nova'`

lentzi90:
> @lentzi90 with single quote around the echo,
> `echo 'export OPENSTACK_SSH_AUTHORIZED_KEY='"'$(cat $CAPO_E2E_SSH_KEY)'"''`
> Gives
> `export OPENSTACK_SSH_AUTHORIZED_KEY='ssh-rsa .....Generated-by-Nova'`

This looks fine to me. :slightly_smiling_face:

Xenwar:
There are some significant changes that make this PR outdated. https://github.com/Nordix/airship-dev-tools/pull/320

Xenwar:
And, Upon closer inspection, the cirros user described in the title is no longer the problem. Regardless of the user, cirros or ubuntu, were not successful in running the e2e test.

PR #324, opened 2021-06-17
Uplift to k8s 1.21.2

kashifest:


lentzi90 (commented):
LGTM!

I found a reference to v1.21.1 that did not change, but not sure if it is relevant. It seems to be some kind of workaround for etcd, so maybe it is fixed in v1.21.2?
https://github.com/Nordix/airship-dev-tools/blob/44159f07cc453f8907c39559695849e800e532a3/ci/scripts/image_scripts/provision_node_image_centos.sh#L30

kashifest:
> LGTM!
> 
> I found a reference to v1.21.1 that did not change, but not sure if it is relevant. It seems to be some kind of workaround for etcd, so maybe it is fixed in v1.21.2?
> https://github.com/Nordix/airship-dev-tools/blob/44159f07cc453f8907c39559695849e800e532a3/ci/scripts/image_scripts/provision_node_image_centos.sh#L30

Yeah, its not needed anymore since the etcd patch has landed in this version.

lentzi90 (approved):

macaptain (approved):

PR #329, opened 2021-07-09
Replace hardcoded values

Xenwar:
Replace hardcoded values

### DIFF
--- a/capo/configure.sh
+++ b/capo/configure.sh
@@ -57,7 +57,6 @@ clouds:
   "${OS_CLOUD_NAME}":
     auth:
       auth_url: "${OS_AUTH_URL}"
-      project_name: Default Project 37137
       username: "${OS_USERNAME}"
       password: "${OS_PASSWORD}"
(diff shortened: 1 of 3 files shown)


capo/configure.sh
```diff
@@ -57,7 +57,6 @@ clouds:
   "${OS_CLOUD_NAME}":
     auth:
       auth_url: "${OS_AUTH_URL}"
-      project_name: Default Project 37137
```
lentzi90, line 60: Do we not need this? Would it make sense to parametrize it instead?
```
project_name: "${PROJECT_NAME}"

Xenwar: We do and we still have it at line 66. removed the duplicate one

Xenwar: And, the variable names are in the format `OS_`

lentzi90: Ah thanks for pointing it out! :slightly_smiling_face:

lentzi90 (approved):

PR #333, opened 2021-07-27
Add k8s and kindest/node config to renovate.json

Sunnatillo:


renovate.json
```diff
@@ -1,51 +1,86 @@
...
-    },
-    {
-      "fileMatch": ["^ci/.*$"],
-      "matchStrings": ["CRICTL_VERSION:-\"(?<currentValue>v.*?)\"}"],
-      "depNameTemplate": "kubernetes-sigs/cri-tools",
-      "datasourceTemplate": "github-tags"
+     "datasources": ["docker"],
+     "packageNames": ["kindest/node"],
+     "automerge": false,
+     "commitMessageTopic": "kindest/node"
```
furkatgofurov7, line 15: This looks good, question though, why do we need to define kindest/node here and also inside the  `regexManagers`?

Sunnatillo: Renovate opening PR with name Node.js if it is not defined in package rules. So, I needed to add "commitMessageTopic" value. That`s why it is defined in two places.

renovate.json
```diff
@@ -1,51 +1,86 @@
...
+       "depNameTemplate": "kubernetes/kubernetes",
+       "datasourceTemplate": "github-releases"
+     },
+     {
+       "fileMatch": ["^ci/.*$"],
+       "matchStrings": ["KIND_NODE_IMAGE_VERSION:-\"(?<currentValue>.*?)\"}"],
+       "depNameTemplate": "kindest/node",
+       "datasourceTemplate": "docker"
+     }
+   ],
```
furkatgofurov7, line 72: Also, seems you are stripping down the people from the list a bit, please check again and make sure everyone is in the list

renovate.json
```diff
@@ -1,51 +1,86 @@
...
+     "lentzi90",
+     "macaptain",
+     "mboukhalfa",
+     "namnx228",
+     "Rozzii",
+     "smoshiur1237",
+     "Sunnatillo",
+     "wgslr",
+     "Xenwar"
+   ],
```
furkatgofurov7, line 84: Also, seems you are stripping down the people from the list a bit, please check again and make sure everyone is in the list

ci/scripts/image_scripts/source_cluster_container_images.sh
```diff
@@ -10,7 +10,8 @@ export IMG_CENTOS_IMG="${IMG_CENTOS_IMG:-"docker.io/centos:centos8"}"
 
 if [[ "${IMAGE_OS}" == "Ubuntu" ]]; then
     export KUBERNETES_VERSION=${KUBERNETES_VERSION:-"v1.21.2"}
-    export IMG_KIND_NODE_IMAGE="${IMG_KIND_NODE_IMAGE:-"kindest/node:v1.21.1"}"
+    export KIND_NODE_IMAGE_VERSION=${KIND_NODE_IMAGE_VERSION:-"v1.21.1"}
```
fmuyassarov, line 13: you could also rely on `KUBERNETES_VERSION` var from L12, because it holds the same value as in `KIND_NODE_IMAGE_VERSION`

fmuyassarov (approved):

ci/scripts/image_scripts/source_cluster_container_images.sh
furkatgofurov7: No, `KIND_NODE_IMAGE_VERSION` is N-1 behind from `KUBERNETES_VERSION` (v1.21.2 & v1.221.1)

fmuyassarov: here yes, but in metal3-dev-env they are the same. And the reason they are not the same here, is because Renovate didn't update the version of Kind yet.

furkatgofurov7: even in m3-dev-env it is N-1 version behind.

fmuyassarov: check here:
https://github.com/metal3-io/metal3-dev-env/blob/ee6f13217c81736385d49ba1c23e0b50265f495a/lib/common.sh#L202
https://github.com/metal3-io/metal3-dev-env/blob/ee6f13217c81736385d49ba1c23e0b50265f495a/lib/common.sh#L220

renovate.json
Sunnatillo: I added

ci/scripts/image_scripts/source_cluster_container_images.sh
furkatgofurov7: No, this is because yesterday renovate updated kind node image to v1.21.2 and we merged it but k8s PR is still there which is updated to v1.21.3 and not merged waiting for this PR to be merged. Check these:
[kind image PR](https://github.com/metal3-io/metal3-dev-env/pull/740)
[k8s PR](https://github.com/metal3-io/metal3-dev-env/pull/739)

furkatgofurov7 (commented):
/lgtm

ci/scripts/image_scripts/source_cluster_container_images.sh
fmuyassarov: why is it behind?

furkatgofurov7: no idea

renovate.json
namnx228: Why don't we need the `commitMessageTopic` for other packages, but only for `kindest/node`?

namnx228 (changes requested):
Overall, it looks good to me. Just have a couple of notices. 
- Please avoid to add a space in the head of every line in `renovate.json`. From the github point of view, it is like you made change to the whole file, so Github highlight...everything in that file. It would make easier for reviewers if only the change that you added is highlight. 
- I also left a comment in line 15

renovate.json
Sunnatillo: Without commitMessageTopic renovate opening a PR with name Update node.js to xx.xx. So I needed to specify kindest/node in Package rules. Other package names were correct when renovate opened a PR.

PR #339, opened 2021-08-02
Revert calico images to v3.19.1

fmuyassarov:
Calico has released new images 3.20.0 but the latest manifest they have is still referencing to 3.191. We have pre-downloaded 3.20.0 images because they are available but in metal3-dev-env we are using the latest calico yaml which is pointing to 3.19.1 images.


namnx228 (approved):
LGTM

furkatgofurov7 (approved):
/approve

PR #341, opened 2021-08-03
Use latest calico images in k8s 1.21.1 node images

fmuyassarov:
Use latest calico images in k8s 1.21.1 node images. For that we need to update the `KUBERNETES_VERSION` variable so that we can trigger the image building job for 1.21.1 which will pre-download the v3.20.0 calico images.


ci/scripts/image_scripts/source_cluster_container_images.sh
```diff
@@ -9,8 +9,8 @@ export IMG_GOLANG_IMG="${IMG_GOLANG_IMG:-"registry.hub.docker.com/library/golang
 export IMG_CENTOS_IMG="${IMG_CENTOS_IMG:-"docker.io/centos:centos8"}"
 
 if [[ "${IMAGE_OS}" == "Ubuntu" ]]; then
-    export KUBERNETES_VERSION=${KUBERNETES_VERSION:-"v1.21.2"}
-    export KIND_NODE_IMAGE_VERSION=${KIND_NODE_IMAGE_VERSION:-"v1.21.2"}   
+    export KUBERNETES_VERSION=${KUBERNETES_VERSION:-"v1.21.1"}
+    export KIND_NODE_IMAGE_VERSION=${KIND_NODE_IMAGE_VERSION:-"v1.21.1"}
```
furkatgofurov7, line 13: Usually, kind node image should be (k8s_version -1), but in this case, since we already updated m3-dev-env var for it with latest(v1.21.2), thinking if this is okay or we should not touch this at all?

furkatgofurov7: Why I am saying it should be one version lower and why as is now not because there is new k8s version (v1.21.3) but we have not upgraded to it yet, so for now they are same, but that is not the case once we update k8s version.

fmuyassarov: good point. Since this is needed for prepulling the kind image, and in metal3-dev-env it's requires 1.21.2 we can keep it as it is (1.21.2) so that when running metal3-dev-env CI tests we don't have to download it and eventually somewhat reduce the possibility of reaching docker pull rate limit. I will revert this line now.

furkatgofurov7 (approved):
/approve

Let's revert these back once we finish building node images, thanks.

PR #349, opened 2021-08-17
Implement scripts to build centos based IPA

Rozzii:


ci/scripts/image_scripts/build_ipa_on_centos.sh
```diff
@@ -0,0 +1,60 @@
...
+IPA_BUILD_WORKSPACE="${IPA_BUILD_WORKSPACE:-/tmp/dib}"
+IPA_REPO="${IPA_REPO:-https://github.com/Nordix/ironic-python-agent}"
+IPA_REPO_CONF="${IPA_IMAGE_NAME} git /tmp/ironic-python-agent ${IPA_REPO}"
+IPA_REPO_CONF_PATH="${IPA_BUILDER_PATH}/dib/ironic-python-agent-ramdisk/source-repository-ironic-python-agent"
+RT_URL="${RT_URL:-https://artifactory.nordix.org/artifactory}"
+# Install required packages
+sudo apt install --yes python3-pip  python3-virtualenv qemu-utils
+
+# Create the work directory
+mkdir "${IPA_BUILD_WORKSPACE}"
```
fmuyassarov, line 22:
```suggestion
mkdir -p "${IPA_BUILD_WORKSPACE}"
```
In case we change the directory to be non-existing subdirectory

ci/scripts/image_scripts/build_ipa_on_centos.sh
```diff
@@ -0,0 +1,60 @@
...
+# IDENTIFIER is the git commit of the HEAD and the ISO 8061 UTC timestamp
+IPA_IDENTIFIER="$(date --utc +"%Y-%m-%dT%H:%MZ")-$(git rev-parse HEAD)"
+cd ..
+
+# Install the cloned IPA builder tool
+virtualenv myenv-opensuse
+# shellcheck source=/dev/null
+source "./myenv-opensuse/bin/activate"
+python3 -m pip install --upgrade pip
+python3 -m pip install "${IPA_BUILDER_PATH}"
```
fmuyassarov, line 43: if we are using pip to install ironic-python-agent-builder, why do we need to clone it in L26?

fmuyassarov (commented):
I haven't reviewed yet the CI related part. Will take a look later on today.

ci/scripts/image_scripts/build_ipa_on_centos.sh
Rozzii: I clone it separately to have easy access to git metadata.

kashifest (commented):
Looks good in the first glance. Some comments inline.

ci/scripts/image_scripts/start_centos_ipa_build.sh
```diff
@@ -0,0 +1,103 @@
...
+  echo "Creating a test executer volume from copy of base volume."
+  create_build_executer_volume "${BASE_VOLUME_NAME_ID}" "${BUILDER_VM_NAME}"
+
+  # Wait for a test executer volume to be available
+  echo "Waiting for a test executer volume to be available."
+  wait_for_volume "${BASE_VOLUME_NAME_ID}" "${BUILDER_VM_NAME}"
+
+  # Resize test executer volume
+  echo "Resizing test executer volume."
+  openstack volume set --size "${RESIZED_VM_SIZE}" "${BUILDER_VM_NAME}"
```
kashifest, line 51: Would we want to use volume for IPA building?

ci/scripts/image_scripts/build_ipa_on_centos.sh
```diff
@@ -0,0 +1,61 @@
...
+# Install the cloned IPA builder tool
+virtualenv venv
+# shellcheck source=/dev/null
+source "./venv/bin/activate"
+python3 -m pip install --upgrade pip
+python3 -m pip install "${IPA_BUILDER_PATH}"
+
+# Build the IPA initramfs and kernel images
+ironic-python-agent-builder --output "${IPA_IMAGE_NAME}" \
+    --release 8 --branch "master" centos --verbose
```
kashifest, line 47: I would rather parameterize the OS here

kashifest:
Shall we think about the naming i.e. shall we keep it `centos` or generalize and parameterize OS inline?

kashifest:
One more thing, I am thinking shall we copy the VM builder scripts or shall we trigger it from project-infra as like other jobs?

ci/scripts/image_scripts/utils.sh
```diff
@@ -0,0 +1,138 @@
...
+  echo "${1:?}-subnet"
+}
+
+# Description:
+# Waits for SSH connection to come up for a server
+#
+# Usage:
+#   wait_for_ssh <ssh_user> <ssh_key_path> <server>
+#
+wait_for_ssh() {
```
fmuyassarov, line 29: you don't have to write it, we already have this  function in airship-dev-tools
https://github.com/Nordix/airship-dev-tools/blob/10470a95fd83b7e52d85e2622e52e0beb45b6a81/ci/scripts/openstack/utils.sh#L541

ci/scripts/image_scripts/utils.sh
```diff
@@ -0,0 +1,138 @@
...
+# Generates subnet name from Network name.
+#
+# Example:
+#   Input: "airship-network"
+#   Output: "airship-network-subnet"
+#
+get_subnet_name() {
+  echo "${1:?}-subnet"
+}
+
```
fmuyassarov, line 22: same here, this function already exists in airship-dev-tools repo
https://github.com/Nordix/airship-dev-tools/blob/10470a95fd83b7e52d85e2622e52e0beb45b6a81/ci/scripts/openstack/utils.sh#L101

fmuyassarov:
> One more thing, I am thinking shall we copy the VM builder scripts or shall we trigger it from project-infra as like other jobs?
It would be nice to reuse the scripts from project-infra repo, but there are more things in that scripts which are not needed for building IPA images. But I have a comment a bit related to this. See https://github.com/Nordix/airship-dev-tools/pull/349#discussion_r690990157

ci/scripts/image_scripts/build_ipa_on_centos.sh
Rozzii: Yes I think we can use this as a general purpose IPA builder that can build IPA on multiple distro, I don't see any problem.

ci/scripts/image_scripts/start_centos_ipa_build.sh
Rozzii: No, it will be removed.

Rozzii:
Thanks for the reviews. I have removed the volumes related code and modified the IPA builder script to accept parameters for "base os" and "base os release".

kashifest:
Shall we rename the file names as well to make it generic @Rozzii

ci/scripts/image_scripts/build_ipa_on_centos.sh
fmuyassarov: I see, so it is for 
`IPA_IDENTIFIER="$(date --utc +"%Y-%m-%dT%H:%MZ")-$(git rev-parse HEAD)"`. Makes sense.

ci/scripts/image_scripts/start_centos_ipa_build.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+# Fail the script if any command fails
+set -eu
+
+# VM configuration variables
+RESIZED_VM_SIZE="${RESIZED_VM_SIZE:-100}"
+VM_TIMELABEL="${VM_TIMELABEL:-$(date '+%Y%m%d%H%M%S')}"
+BUILDER_VM_NAME="${BUILDER_VM_NAME:-ci-builder-vm-${VM_TIMELABEL}}"
+BUILDER_PORT_NAME="${BUILDER_PORT_NAME:-${BUILDER_VM_NAME}-int-port}"
+BUILDER_FLAVOR="${BUILDER_FLAVOR:-2C-8GB}"
```
fmuyassarov, line 10: Nit: just curious if 2 cores is enough. Maybe we can decrease the disk size to 50GB and increase the cores to 4? Any thoughts?
Example: `4Cores, 8GB RAM, 50GB disk `
/cc @kashifest

ci/scripts/image_scripts/build_ipa_on_centos.sh
```diff
@@ -0,0 +1,63 @@
...
+# Install the cloned IPA builder tool
+virtualenv venv
+# shellcheck source=/dev/null
+source "./venv/bin/activate"
+python3 -m pip install --upgrade pip
+python3 -m pip install "${IPA_BUILDER_PATH}"
+
+# Build the IPA initramfs and kernel images
+ironic-python-agent-builder --output "${IPA_IMAGE_NAME}" \
+    --release "${IPA_BASE_OS_RELEASE}" --branch "master" "${IPA_BASE_OS}" --verbose
```
fmuyassarov, line 49:
```suggestion
ironic-python-agent-builder --output "${IPA_IMAGE_NAME}" \
    --release "${IPA_BASE_OS_RELEASE}" -e dynamic-login -e journal-to-console -e devuser -e openssh-server --branch "master" "${IPA_BASE_OS}" --verbose
```
Can you please add these elements as per the request of @maelk ? If I remember correct,  we can't use -e flag once and give a list of elements with comma in between. Maybe we can try that by surrounding the list with double quotes. If not the last option, as an example I give above.

fmuyassarov (commented):
This is looks nice. We are almost there. Left some comments.

ci/scripts/image_scripts/start_centos_ipa_build.sh
Rozzii: We don't really compile anything, basically we just copy files around as far as I understand and ofcourse a bit of downloading and packaging but no computationally heavy thing. I have chosen a the 8GB memory because we utilize tmpfs during the build process and that requires quiet a lot of memory but cpu wise it doesn't seem to be heavy. We can ofcourse change it any time.

fmuyassarov: Then let's keep it as it is for now

ci/scripts/image_scripts/build_ipa_on_centos.sh
Rozzii: I have added these to the script, I have tested the -e option and it looks like we have to use multiple
 -e options to pass multiple additional element names.

ci/scripts/image_scripts/start_centos_ipa_build.sh
kashifest: Agreed with @Rozzii , lets keep it synced with requirements for now

ci/scripts/image_scripts/build_ipa_on_centos.sh
fmuyassarov: ok, that should be fine.

ci/scripts/image_scripts/build_ipa_on_centos.sh
```diff
@@ -0,0 +1,65 @@
...
+# Create the work directory
+mkdir --parents "${IPA_BUILD_WORKSPACE}"
+cd "${IPA_BUILD_WORKSPACE}"
+
+# Pull IPA builder repository
+git clone --single-branch --branch "${IPA_BUILDER_BRANCH}" "${IPA_BUILDER_REPO}"
+
+# Change the IPA repository address that is used during the build process
+echo -n "${IPA_REPO_CONF}" > "${IPA_REPO_CONF_PATH}" 
+
```
fmuyassarov, line 33: I'm not sure I understand what we are doing here and with env vars like `IPA_REPO_CONF` and `IPA_REPO_CONF_PATH`. Can you please elaborate more on this?

fmuyassarov (changes requested):
Thanks for addressing the comments. I have last comments inline. Also, we need to ensure that the size doesn't increase dramatically. Can we add a size check once we have built the IPA and give some kind of signal it is bigger than expected. I'm not sure what is the maximum allowed size, but in [IPA Builder Introduction talk](https://youtu.be/1L1Ld7skgDw?t=1014), people were saying 600 MB was still fine. So, maybe we can set this number as the threshold or even ~500MB ?

fmuyassarov:
@Rozzii One thing I forgot to mention, it would be helpful for someone working on these scripts if we have a document explaining the process in short, even though the script is already self-explanatory. It is not a blocker and can be done separately on a follow up PR.

Rozzii (commented):
Sure @fmuyassarov , I can make a README  with some higher level explanation about the steps, I agree that it could be useful to understand the build process.

Rozzii:
Sure @fmuyassarov , I can make a README  with some higher level explanation about the steps, I agree that it could be useful to understand the build process.

ci/scripts/image_scripts/build_ipa_on_centos.sh
Rozzii: This is basically a patching process, it injects the address of the IPA git repo to the IPA builder. Same thing could be achieved using our own branch on the Nordix IPA fork but it is so minor that I thought a simple patch like this is enough no need to maintain a branch.

ci/scripts/image_scripts/build_ipa_on_centos.sh
```diff
@@ -0,0 +1,65 @@
...
+# IDENTIFIER is the git commit of the HEAD and the ISO 8061 UTC timestamp
+IPA_IDENTIFIER="$(date --utc +"%Y-%m-%dT%H:%MZ")-$(git rev-parse HEAD)"
+cd ..
+
+# Install the cloned IPA builder tool
+virtualenv venv
+# shellcheck source=/dev/null
+source "./venv/bin/activate"
+python3 -m pip install --upgrade pip
+python3 -m pip install "${IPA_BUILDER_PATH}"
```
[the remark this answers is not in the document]
Rozzii: I have used a parameter because this IPA_BUILDER_PATH is present in a other long path and I wanted to shorten it and also it used 3 times so I thought It is easier to change it in the future if it is a variable not a constant. If you'd like it to be a constant I am happy to change it.

After your comment I have modified  it a bit, now it can't be configured with environment variable.

fmuyassarov (commented):
I'm happy with this state. Last question I have, if we need to modify some files of ironic-python-agent-builder repository, for example if we need to add a new package in [pacakge-install.yaml](https://github.com/openstack/ironic-python-agent-builder/blob/master/dib/ironic-python-agent-ramdisk/package-installs.yaml), how we are going to make sure that `ironic-python-agent-builder` utility runs our version of `ironic-python-agent-builder ` code and not the one it finds in upstream git?

fmuyassarov:
> I'm happy with this state. Last question I have, if we need to modify some files of ironic-python-agent-builder repository, for example if we need to add a new package in [pacakge-install.yaml](https://github.com/openstack/ironic-python-agent-builder/blob/master/dib/ironic-python-agent-ramdisk/package-installs.yaml), how we are going to make sure that `ironic-python-agent-builder` utility runs our version of `ironic-python-agent-builder ` code and not the one it finds in upstream git?

The reason I ask is because we are installing the `ironic-python-image-builder` via pip.

ci/scripts/image_scripts/build_ipa_on_centos.sh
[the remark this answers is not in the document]
fmuyassarov: Let's keep it as it is then. It is not a big deal anyways, I was just curious.

Rozzii:
So, we are installing it via pip but we use pip as a local package manager to run the setup.py and other scripts that are part of the installation process. We are not downloading ironic-python-image-builder from a pip repository. We just clone then we execute the python installer script.

fmuyassarov:
> We are installing it via pip but we use pip as a local package manager to run the setup.py and other scripts that are part of the installation process. We are not downloading ironic-python-image-builder from a pip repository. We just clone then we execute the python installer script.

Ah got it. Now I understand that var better. Thank you.

fmuyassarov (dismissed):
@kashifest I'm leaving it to you to merge it so that you can give your final review too. Thanks.

Rozzii:
I will still push a min or modification.

Rozzii:
I have changed the way how the custom repository addresses are handled, there is a slight conflict in among the IPA-builder and the Diskimage-builder. IPA-builder ties together the git refs of the openstack-requirements and the IPA repos by default so they have to be handled separately otherwise any IPA build that tries to do a build based on a arbitrary IPA commit would fail.

fmuyassarov (approved):

PR #355, opened 2021-08-23
Fix misplaced quotation mark

Rozzii:


ci/scripts/image_scripts/start_centos_ipa_build.sh
```diff
@@ -37,9 +37,9 @@ wait_for_ssh "${AIRSHIP_CI_USER}" "${AIRSHIP_CI_USER_KEY}" "${BUILDER_IP}"
 # Send Remote script to Executer
 scp \
   -o StrictHostKeyChecking=no \
-  -o UserKnownHostsFile=/dev/null \
+  -o UserKnownHostsFile="/dev/null" \
```
fmuyassarov, line 40: Is this needed here? Because we are not adding "" quotes in project-infra, and even in the ssh step below

Rozzii: No, it was just highlighted earlier by shellcheck but the error was related to a different problem. I will remove it.

fmuyassarov (approved):

PR #356, opened 2021-08-23
Fix incorrect var in IPA building

fmuyassarov:
Fix incorrect var in IPA building

PR #357, opened 2021-08-23
Fix IPA identifier value

Rozzii:


Rozzii:
I think this has to go in without the ipa test, as https://github.com/Nordix/airship-dev-tools/pull/356 is still waiting to be merged. 
So I think it should go in, an then we can do the test on 356.

fmuyassarov:
> I think this has to go in without the ipa test, as #356 is still waiting to be merged.
> So I think it should go in, an then we can do the test on 356.

True. Just wanted to check. But maybe I should do it on another PR.

fmuyassarov (approved):

PR #360, opened 2021-08-23
Add script to build custom Ironic images

fmuyassarov:
Add script to build custom Ironic images


ci/scripts/image_scripts/build_ironic.sh
```diff
@@ -0,0 +1,25 @@
...
+cd ironic
+IRONIC_TAG="$(git rev-parse --short HEAD)"
+echo "IRONIC_TAG is: ${IRONIC_TAG}"
+cd ..
+
+git clone https://github.com/metal3-io/ironic-image.git
+cd ironic-image
+docker build -t fmuyassarov/ironic-image:"${IRONIC_TAG}" .
+# sudo docker login to where?
+docker push fmuyassarov/ironic-image:${IRONIC_TAG}
```
fmuyassarov, line 24: Where shall we push the images @kashifest ?

fmuyassarov:
Please note: we need to merge https://gerrit.nordix.org/c/infra/cicd/+/9978 first.

Makefile
```diff
@@ -144,6 +144,10 @@ integration_test: ## Run integration test
 build_ipa:
 	$(CURDIR)/ci/scripts/image_scripts/start_centos_ipa_build.sh
 
+.PHONY: build_ironic
+make build_ironic:
```
kashifest, line 148: Remove `make`
```suggestion
build_ironic:
```

ci/jobs/openstack_ipa_and_ironic_image_building.pipeline
```diff
@@ -65,6 +65,21 @@ pipeline {
...
       }
     }
+    stage('Building Ironic container image'){
+      options {
+        timeout(time: 60, unit: 'MINUTES')
+      }
+      steps {
+        withCredentials([usernamePassword(credentialsId: 'airshipci_city_cloud_openstack_credentials', usernameVariable: 'OS_USERNAME', passwordVariable: 'OS_PASSWORD')]) {
+          withCredentials([sshUserPrivateKey(credentialsId: 'airshipci_city_cloud_ssh_keypair', keyFileVariable: 'AIRSHIP_CI_USER_KEY')]) {
+            withCredentials([usernamePassword(credentialsId: 'infra-nordix-artifactory-api-key', usernameVariable: 'RT_USER', passwordVariable: 'RT_TOKEN')]) {
```
kashifest, line 75: I don't think this is needed here?

ci/scripts/image_scripts/build_ironic.sh
```diff
@@ -0,0 +1,25 @@
...
+cd
```
kashifest, line 25: `cd ..`? May be `pushd` and `popd` if you want to always come back to root dir?

kashifest (changes requested):

ci/jobs/openstack_ipa_and_ironic_image_building.pipeline
fmuyassarov: True. Removed

ci/scripts/image_scripts/build_ironic.sh
fmuyassarov: Sure. Switched to pushd/popd

ci/scripts/image_scripts/run_build_ironic.sh
```diff
@@ -0,0 +1,25 @@
...
+IRONIC_TAG="$(git rev-parse --short HEAD)"
+echo "IRONIC_TAG is: ${IRONIC_TAG}"
+popd
+
+git clone https://github.com/metal3-io/ironic-image.git
+pushd ironic-image
+docker build -t fmuyassarov/ironic-image:"${IRONIC_TAG}" .
+# sudo docker login to where?
+docker push fmuyassarov/ironic-image:${IRONIC_TAG}
+popd
```
Rozzii, line 25: No new line at the end of file.

ci/scripts/image_scripts/run_build_ironic.sh
```diff
@@ -0,0 +1,25 @@
...
+# Requered parameter is refspec, which is gerrit refspec of the patch
+# Example: changes/74/804074/1
+
+REFSPEC="${1}"
+echo "The Ironic patch we are going to switch to is: ${REFSPEC}"
+project_dir="openstack/ironic"
+
+git clone https://opendev.org/openstack/ironic.git
+pushd ironic
+IRONIC_TAG="$(git rev-parse --short HEAD)"
```
Rozzii, line 16: I think the timestamp and short git hash combination would be beneficial here too.

Rozzii (changes requested):

ci/scripts/image_scripts/run_build_ironic.sh
fmuyassarov: agreed

Rozzii (dismissed):
There is still this question: https://github.com/Nordix/airship-dev-tools/pull/360#discussion_r693994857 but other than that it looks okay to me.

fmuyassarov:
> There is still this question: [#360 (comment)](https://github.com/Nordix/airship-dev-tools/pull/360#discussion_r693994857) but other than that it looks okay to me.

I have received credentials and added those Jenkins credentials to the pipeline and the script.

ci/scripts/image_scripts/run_build_ironic.sh
```diff
@@ -0,0 +1,37 @@
...
+IMAGE_REGISTRY="registry.nordix.org"
+IMAGE_REPO="airship"
+DOCKER_USER="${DOCKER_USER}"
+DOCKER_PASSWORD="${DOCKER_PASSWORD}"
+docker login "${IMAGE_REGISTRY}" -u $DOCKER_USER -p $DOCKER_PASSWORD
+
+# Create a unique tag for Ironic image
+git clone https://opendev.org/openstack/ironic.git
+pushd ironic
+IRONIC_TAG="$(date --utc +"%Y%m%dT%H%MZ")-$(git rev-parse --short HEAD)"
```
Rozzii, line 24: Let's not version it here, let's create the image version in the Jenkins pipeline script before the build stage would start.
This tag information will be needed by the IPA build test later.

ci/jobs/openstack_ipa_and_ironic_image_building.pipeline
```diff
@@ -65,6 +65,17 @@ pipeline {
...
+      options {
+        timeout(time: 30, unit: 'MINUTES')
+      }
+      steps {
+          withCredentials([usernamePassword(credentialsId: 'airshipci_harbor', usernameVariable: 'DOCKER_USER', passwordVariable: 'DOCKER_PASSWORD')])  {
+              /* Ironic image building script */
+              sh "make build_ironic"
+          }
+      }   
+    }
```
Rozzii, line 78: I think we should have it before the IPA building stage, so this should be the first stage.

ci/scripts/image_scripts/run_build_ironic.sh
```diff
@@ -0,0 +1,37 @@
...
+# build & push the Ironic container image 
+git clone https://opendev.org/openstack/
+git clone https://github.com/metal3-io/ironic-image.git
+pushd ironic-image
+sudo docker build -t "${IMAGE_REPO}/ironic-image:${IRONIC_TAG}" .
+sudo docker push "${IMAGE_REPO}/ironic-image:${IRONIC_TAG}"
+popd
+
+# Logout from the Nordix container registry
+docker logout "${IMAGE_REGISTRY}"
```
Rozzii, line 37: Empty line before EOF is missing.

Rozzii (changes requested):
I think this has to be a parameterized build, so the tag of the ironic image has to be generated from the combination of the timestamp and the parameter that is containing the ironic commit hash. Later the during the IPA testing we will pull the image that has been built during the first stage.

fmuyassarov:
> I think this has to be a parameterized build, so the tag of the ironic image has to be generated from the combination of the timestamp and the parameter that is containing the ironic commit hash. Later the during the IPA testing we will pull the image that has been built during the first stage.

We can't include the timestamp because whenever we build the image it will have a unique tag. As such, there is no way to really verify if the image exists already on the registry or not. That's why I excluded timestamp part.

ci/scripts/image_scripts/run_build_ironic.sh
fmuyassarov: I would like to avoid adding shell scripts within Jenkins pipeline because there are couple vars we need to inherit for creating a tag. I am checking how to pass the var from the current shell to the next Jenkins pipeline stage.

ci/scripts/image_scripts/run_build_ironic.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+set -eu
+
+# This script relies on usptream https://github.com/metal3-io/ironic-image/blob/master/patch-image.sh
```
kashifest, line 5: upstream

ci/scripts/image_scripts/run_build_ironic.sh
```diff
@@ -0,0 +1,56 @@
...
+# to build Ironic container image based on a gerrit refspec of a patch.
+# Requered parameter is REFSPEC, which is gerrit refspec of the patch
```
kashifest, line 7: Required

kashifest (approved):
Two small nits

ci/scripts/image_scripts/run_build_ironic.sh
fmuyassarov: ops, thanks

ci/jobs/openstack_ipa_and_ironic_image_building.pipeline
```diff
@@ -76,4 +87,4 @@ pipeline {
       }
     }
   }
-}
+}
```
Rozzii: Missing newline at EOF.

Rozzii (commented):
You can add multiple tag for an OCI image with docker e.g. docker build -t name1:tag1 -t name1:tag2 so you can use both timestamp and git hash separately.

ci/scripts/image_scripts/run_build_ironic.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+set -eu
+
+# This script relies on upstream https://github.com/metal3-io/ironic-image/blob/master/patch-image.sh
+# to build Ironic container image based on a gerrit refspec of a patch.
+# Required parameter is REFSPEC, which is gerrit refspec of the patch
+# Example: refs/changes/74/804074/1
+ 
+REFSPEC="${REFSPEC:-}"
```
Rozzii, line 10: In my opinion In the JJB config this parameter needs to be changed to something more descriptive eg.  IRONIC_REFSPEC as the same job will handle ironic and IPA build.

fmuyassarov:
> You can add multiple tag for an OCI image with docker e.g. docker build -t name1:tag1 -t name1:tag2 so you can use both timestamp and git hash separately.

True. But the problem is not being able to add the timestamp, but how do we then make sure if that image exists on the registry or not? Because it will have completely unique tag which won't allow us to verify if it is the same image as build now.

PR #363, opened 2021-08-25
Add basic bootstrap test to IPA build process

Rozzii:
This test simply checks whether the metal3-dev-env could
be bootstrapped by utilizing the newly built IPA archive.

In case the "bootstrap test" fails the IPA archive won't
be uploaded to Nordix artifactory.


Rozzii:
This PR can be merged after  both:
https://github.com/metal3-io/baremetal-operator/pull/957
and
https://github.com/metal3-io/metal3-dev-env/pull/774
are already merged.

PR #368, opened 2021-08-31
Add build artifact cleanup logic to pipeline

Rozzii:
This commit adds a Artifactory directory cleanup logic to
ci/scripts/artifactory/utils.sh. The general purpose cleanup logic
supports artifact pinning (exclusion from cleanup), dry-run and
artifact retention functionalities.

This commit also creates a pipeline script and related shell scripts
to provide logic for a job to clean any artifact directory.

At the moment the cleanup pipeline only check centos stream based
IPA build artifacts but that can be extended by adding additional
Artifactory urls to the script.

### DIFF
--- a/ci/scripts/image_scripts/artifactory_cleanup.sh
+++ b/ci/scripts/image_scripts/artifactory_cleanup.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+# The goal of this script is to clean artifacts from specified Artifactory directories.
+
+set -eu
+
+# General artifactory variables
+RT_UTILS="${RT_UTILS:-/tmp/utils.sh}"
+RT_URL="https://artifactory.nordix.org/artifactory"
+IPA_ROOT_ARTIFACTORY="airship/images/ipa"
+DRY_RUN="${DRY_RUN:-false}"
+ANONYM="${ANONYM:-0}"
+
+# shellcheck disable=SC1090
+source "${RT_UTILS}"
+export "RT_URL=${RT_URL}"
+
+# Centos Stream configuration
+CENTOS_STREAM_ROOT="${IPA_ROOT_ARTIFACTORY}/centos/stream"
+
+# Centos Stream Review
+CENTOS_STREAM_REVIEW="${CENTOS_STREAM_ROOT}/review"
+CENTOS_STREAM_REVIEW_PINNED="${PINNED_ARTIFACTS:-/tmp/centos_stream_review_pinned.txt}"
+CENTOS_STREAM_REVIEW_RETENTION_NUM="${CENTOS_STREAM_REVIEW_RETENTION_NUM:-5}"
+
+rt_delete_multiple_artifacts "${CENTOS_STREAM_REVIEW}" "${ANONYM}" "${DRY_RUN}" \
+    "${CENTOS_STREAM_REVIEW_PINNED}" "${CENTOS_STREAM_REVIEW_RETENTION_NUM}"
+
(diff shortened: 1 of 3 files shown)


ci/scripts/image_scripts/start_artifactory_cleanup.sh
```diff
@@ -0,0 +1,59 @@
...
+ssh \
+  -o StrictHostKeyChecking=no \
+  -o UserKnownHostsFile=/dev/null \
+  -o ServerAliveInterval=15 \
+  -o ServerAliveCountMax=10 \
+  -i "${AIRSHIP_CI_USER_KEY}" \
+  "${AIRSHIP_CI_USER}"@"${IP}" \
+  PATH=/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/sbin:/bin \
+  RT_USER="${RT_USER}" RT_TOKEN="${RT_TOKEN}" "/tmp/${CLEANUP_SCRIPT_NAME}"
+
```
fmuyassarov, line 59: One alternative to executing the script on a temporary VM would be running it within a temporary container on the Jenkins node. I see that we  do such executions for node image [building jobs](https://github.com/Nordix/airship-dev-tools/blob/bbeec9fdf09fe8f19365a2e973faa5cc6ffeaad5/ci/jobs/openstack_node_image_building.pipeline#L57-L69). Since it is within a container there is nothing effecting the host, and `--rm` ensures container is deleted by the end of its run.  I'm OK with VM too,  but using container would avoid creation and deletion scripts. But I let you decide

ci/scripts/image_scripts/artifactory_cleanup.sh
```diff
@@ -0,0 +1,33 @@
...
+CENTOS_STREAM_REVIEW="${CENTOS_STREAM_ROOT}/review"
+CENTOS_STREAM_REVIEW_PINNED="${PINNED_ARTIFACTS:-/tmp/centos_stream_review_pinned.txt}"
+CENTOS_STREAM_REVIEW_RETENTION_NUM="${CENTOS_STREAM_REVIEW_RETENTION_NUM:-5}"
+
+rt_delete_multiple_artifacts "${CENTOS_STREAM_REVIEW}" "${ANONYM}" "${DRY_RUN}" \
+    "${CENTOS_STREAM_REVIEW_PINNED}" "${CENTOS_STREAM_REVIEW_RETENTION_NUM}"
+
+# Centos Stream Staging
+CENTOS_STREAM_STAGING="${CENTOS_STREAM_ROOT}/staging"
+CENTOS_STREAM_STAGING_PINNED="${PINNED_ARTIFACTS:-/tmp/centos_stream_staging_pinned.txt}"
```
fmuyassarov, line 28: if I understood it right, this file contains directories in the artifactory that we should not delete. I wonder why we need it here if it is empty

fmuyassarov (commented):
Would you mind to fix issue in L99 too please?

ci/scripts/image_scripts/artifactory_cleanup.sh
Rozzii: I just put it here to make it faster for the next person who will pin some artifact, it is just a placeholder now.

fmuyassarov (approved):

ci/scripts/image_scripts/start_artifactory_cleanup.sh
Rozzii: Process has been changed to use docker.

PR #374, opened 2021-09-03
Fix missing make command in build_ipa.sh

Rozzii:


ci/scripts/image_scripts/build_ipa.sh
```diff
@@ -72,6 +72,11 @@ tar --create --verbose --file="${IPA_IMAGE_TAR}" \
     "./${IPA_IMAGE_NAME}.kernel" \
     "./${IPA_IMAGE_NAME}.initramfs"
 
+wget "https://golang.org/dl/go1.17.linux-amd64.tar.gz"
+rm -rf "/usr/local/go"
+tar -C "/usr/local -xzf go1.17.linux-amd64.tar.gz"
+export PATH=$PATH:/usr/local/go/bin
+
```
fmuyassarov, line 79: this is redundant I think? or do we need golang before integration tests?

fmuyassarov (changes requested):

ci/jobs/openstack_ipa_and_ironic_image_building.pipeline
```diff
@@ -70,12 +73,21 @@ pipeline {
   }
   post{
     always{
+      script {
+        CURRENT_END_TIME = System.currentTimeMillis()
+        if ((((CURRENT_END_TIME - CURRENT_START_TIME)/1000) - TIMEOUT) > 0) {
```
fmuyassarov, line 78: `TIMEOUT` var is not declared anywhere. You either need to declare it and use that in Line55 or harcode the same value here. I would go for the first option.

fmuyassarov (changes requested):

ci/jobs/openstack_ipa_and_ironic_image_building.pipeline
Rozzii: Done.

Rozzii:
I will add log collection in a separate PR.

ci/jobs/openstack_ipa_and_ironic_image_building.pipeline
```diff
@@ -52,16 +55,19 @@ pipeline {
   stages {
     stage('Building ipa image'){
       options {
-        timeout(time: 60, unit: 'MINUTES')
+        timeout(time: 120, unit: 'MINUTES')
```
fmuyassarov, line 58:
```suggestion
        timeout(time: TIMEOUT, unit: 'MINUTES')
```

fmuyassarov (changes requested):

fmuyassarov (approved):

PR #375, opened 2021-09-06
Archive the logs of the IPA build process

Rozzii:


Rozzii:
This PR will be rebased after https://github.com/Nordix/airship-dev-tools/pull/374 have been merged

namnx228:
LGTM

ci/scripts/image_scripts/run_fetch_logs.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
```
macaptain, line 1: This script has been copied from the project-infra repo. What do you plan to do if the project-infra repo version of the script changes? Would this problem be avoided if your pipelines were in the project-infra repo instead of the airship-dev-tools repo?

ci/scripts/image_scripts/collect_ipa_test_logs.sh
```diff
@@ -0,0 +1,43 @@
+#! /usr/bin/env bash
```
macaptain, line 1: This script is called 'collect_ipa_test_logs.sh'. But is there anything about it specific to IPA? Isn't this 'fetch_logs.sh' copied from project-infra?

Rozzii: No not really I agree a generic name would be better for this.

smoshiur1237: I do have the same opinion to have a generic name.

ci/scripts/image_scripts/run_fetch_logs.sh
Rozzii: We have discussed this offline, this question leads to a bigger question of organizing processes between dev-tools and project infra and that should be resolved outside of this PR.

macaptain (approved):

fmuyassarov (commented):
What if we just run the scripts from project-infra ?

Rozzii:
Well we can do it but then we need extra logic that pulls the project infra source so basically we would need a wrapper and also some workspace cleanup policy because we will pull the project-infra repository to the workspace on the static VM, we can got to that direction too.

fmuyassarov:
> Well we can do it but then we need extra logic that pulls the project infra source so basically we would need a wrapper and also some workspace cleanup policy because we will pull the project-infra repository to the workspace on the static VM, we can go to that direction too.

True. Well maybe let's keep it for now until we come up with something to avoid code duplication.

fmuyassarov (approved):

PR #378, opened 2021-09-08
Fix incorrect IPA cloning process

Rozzii:
With this commit it will be possible to build IPA based on any
commit in the repository.


fmuyassarov (approved):
How come we missed it 🤦‍♂️

PR #398, opened 2021-09-27
Inherit KIND_NODE_IMAGE_VERSION value from JJB

fmuyassarov:
We have added `KIND_NODE_IMAGE_VERSION` to our JJB so that we can manipulate the value while triggering the job. However, this var's value is not inherited in the next scripts because we haven't added it in our jenkins pipeline to fetch it from JJB.


fmuyassarov:
cc @kashifest

kashifest (approved):

PR #399, opened 2021-09-27
Add Ironic and Ironic-inspector vars in ironic image builiding process

smoshiur1237:
This PR will add Ironic and Ironic-inspector vars in image building and export them to metadata file.

### DIFF
--- a/ci/scripts/image_scripts/run_build_ironic.sh
+++ b/ci/scripts/image_scripts/run_build_ironic.sh
@@ -10,8 +10,12 @@ sudo apt update -y
 IMAGE_REGISTRY="registry.nordix.org"
 CONTAINER_IMAGE_REPO="airship"
 IRONIC_REFSPEC="${IRONIC_REFSPEC:-}"
+IRONIC_INSPECTOR_REFSPEC="${IRONIC_INSPECTOR_REFSPEC:-}"
+IRONIC_INSPECTOR_REPO="https://opendev.org/openstack/ironic-inspector.git"
 IRONIC_REPO="https://opendev.org/openstack/ironic.git"
 IRONIC_IMAGE_REPO="https://github.com/metal3-io/ironic-image.git"
+IRONIC_IMAGE_REPO_COMMIT="${IRONIC_IMAGE_REPO_COMMIT:-"HEAD"}"
+IRONIC_IMAGE_BRANCH="${IRONIC_IMAGE_BRANCH:-"master"}"
 PATCH_LIST_FILE=patchList.txt
 
 # Login to the Nordix container registry
@@ -31,6 +35,18 @@ cat << EOF > /tmp/vars.sh
 IRONIC_TAG="${IRONIC_TAG}"
 EOF
 
+# Ironic and Ironic inspector vars are added to metaldata file
+touch /tmp/metadata.txt
+cat << EOF > /tmp/metadata.txt
+IRONIC_REFSPEC="${IRONIC_REFSPEC}"
+IRONIC_REPO="${IRONIC_REPO}"
+IRONIC_IMAGE_REPO="${IRONIC_IMAGE_REPO}"
+IRONIC_IMAGE_REPO_COMMIT="${IRONIC_IMAGE_REPO_COMMIT}"
+IRONIC_IMAGE_BRANCH="${IRONIC_IMAGE_BRANCH}"
+IRONIC_INSPECTOR_REFSPEC="${IRONIC_INSPECTOR_REFSPEC}"
+IRONIC_INSPECTOR_REPO="${IRONIC_INSPECTOR_REPO}"
+EOF
+
 # Build & push the Ironic container image
 # Push image only if it doesn't already exist in the registry
 if docker manifest inspect "${CONTAINER_IMAGE_REPO}/ironic-image:${IRONIC_TAG}"  > /dev/null; then
@@ -39,6 +55,8 @@ else
     echo "${CONTAINER_IMAGE_REPO}/ironic-image:${IRONIC_TAG} image does not exist -> going to push to Harbor"
     git clone ${IRONIC_IMAGE_REPO}
     pushd ironic-image
+    git checkout "${IRONIC_IMAGE_BRANCH}"
+    git checkout "${IRONIC_IMAGE_REPO_COMMIT}"
 
     # Create a patchlist
     touch ${PATCH_LIST_FILE}
@@ -50,6 +68,7 @@ else
       PROJECT_DIR="openstack/ironic"
       cat << EOF | tee -a ${PATCH_LIST_FILE}
 ${PROJECT_DIR} ${IRONIC_REFSPEC}
+${PROJECT_DIR}-inspector ${IRONIC_INSPECTOR_REFSPEC}
 EOF
     # Build container image
       docker build -t "${IMAGE_REGISTRY}/${CONTAINER_IMAGE_REPO}/ironic-image:${IRONIC_TAG}" --build-arg PATCH_LIST=${PATCH_LIST_FILE} .
@@ -61,4 +80,3 @@ fi
 
 # Logout from the Nordix container registry
 docker logout "${IMAGE_REGISTRY}"
-
--- a/ci/scripts/image_scripts/start_centos_ipa_ironic_build.sh
+++ b/ci/scripts/image_scripts/start_centos_ipa_ironic_build.sh
@@ -55,7 +55,11 @@ ssh \
   -i "${AIRSHIP_CI_USER_KEY}" \
   "${AIRSHIP_CI_USER}"@"${BUILDER_IP}" \
   PATH=/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/sbin:/bin \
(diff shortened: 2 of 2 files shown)


ci/scripts/image_scripts/run_build_ironic.sh
```diff
@@ -10,8 +10,13 @@ sudo apt update -y
 IMAGE_REGISTRY="registry.nordix.org"
 CONTAINER_IMAGE_REPO="airship"
 IRONIC_REFSPEC="${IRONIC_REFSPEC:-}"
+IRONIC_INSPECTOR_REFSPEC="${IRONIC_INSPECTOR_REFSPEC:-}"
+IRONIC_INSPECTOR_IMAGE_REPO_REF="${IRONIC_INSPECTOR_IMAGE_REPO_REF:-"HEAD"}"
+IRONIC_INSPECTOR_IMAGE_BRANCH="${IRONIC_INSPECTOR_IMAGE_BRANCH:-"master"}"
```
Rozzii, line 15: IRONIC_INSPECTOR_IMAGE_BRANCH is not in use.
IRONIC_INSPECTOR_IMAGE_REPO_REF is not in use.
IRONIC_IMAGE_BRANCH  is not in use.
If these are actually not used to checkout data then these can be removed from the pipeline as parameters and then we can get their value retrospectively after the repo was cloned.

Rozzii (changes requested):

ci/scripts/image_scripts/run_build_ironic.sh
smoshiur1237: Thanks good point, will update the script.

ci/scripts/image_scripts/run_build_ironic.sh
```diff
@@ -31,6 +36,17 @@ cat << EOF > /tmp/vars.sh
 IRONIC_TAG="${IRONIC_TAG}"
 EOF
 
+# Ironic and Ironic inspector vars are added to metaldata file
+touch /tmp/metadata.txt
+cat << EOF > /tmp/metadata.txt
+IRONIC_REFSPEC="${IRONIC_REFSPEC}"
```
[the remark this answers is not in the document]
smoshiur1237: Sure, will add the necessary repo URL.

Rozzii (approved):

ci/scripts/image_scripts/run_build_ironic.sh
```diff
@@ -10,8 +10,12 @@ sudo apt update -y
 IMAGE_REGISTRY="registry.nordix.org"
 CONTAINER_IMAGE_REPO="airship"
 IRONIC_REFSPEC="${IRONIC_REFSPEC:-}"
+IRONIC_INSPECTOR_REFSPEC="${IRONIC_INSPECTOR_REFSPEC:-}"
+IRONIC_INSPECTOR_REPO="https://opendev.org/openstack/ironic-inspector.git"
```
fmuyassarov, line 14: I think we don't need it? the only usage I see here is adding it in `metadata.txt,` which is still redundant I believe. WDYT @Rozzii ?

ci/scripts/image_scripts/run_build_ironic.sh
```diff
@@ -31,6 +35,18 @@ cat << EOF > /tmp/vars.sh
 IRONIC_TAG="${IRONIC_TAG}"
 EOF
 
+# Ironic and Ironic inspector vars are added to metaldata file
+touch /tmp/metadata.txt
+cat << EOF > /tmp/metadata.txt
+IRONIC_REFSPEC="${IRONIC_REFSPEC}"
+IRONIC_REPO="${IRONIC_REPO}"
+IRONIC_IMAGE_REPO="${IRONIC_IMAGE_REPO}"
```
fmuyassarov, line 43: although it doesn't hurt to add them, probably we don't need them in `metadata.txt`

ci/scripts/image_scripts/run_build_ironic.sh
```diff
@@ -31,6 +35,18 @@ cat << EOF > /tmp/vars.sh
...
+IRONIC_IMAGE_REPO_REF="${IRONIC_IMAGE_REPO_REF}"
+IRONIC_IMAGE_BRANCH="${IRONIC_IMAGE_BRANCH}"
+IRONIC_INSPECTOR_REFSPEC="${IRONIC_INSPECTOR_REFSPEC}"
+IRONIC_INSPECTOR_REPO="${IRONIC_INSPECTOR_REPO}"
```
fmuyassarov, line 47: same here

ci/scripts/image_scripts/run_build_ironic.sh
```diff
@@ -50,6 +68,7 @@ else
       PROJECT_DIR="openstack/ironic"
       cat << EOF | tee -a ${PATCH_LIST_FILE}
 ${PROJECT_DIR} ${IRONIC_REFSPEC}
+${PROJECT_DIR}-inspector ${IRONIC_INSPECTOR_REFSPEC}
```
fmuyassarov, line 71: nice hack :)

ci/scripts/image_scripts/run_build_ironic.sh
```diff
@@ -10,8 +10,12 @@ sudo apt update -y
 IMAGE_REGISTRY="registry.nordix.org"
 CONTAINER_IMAGE_REPO="airship"
 IRONIC_REFSPEC="${IRONIC_REFSPEC:-}"
+IRONIC_INSPECTOR_REFSPEC="${IRONIC_INSPECTOR_REFSPEC:-}"
+IRONIC_INSPECTOR_REPO="https://opendev.org/openstack/ironic-inspector.git"
 IRONIC_REPO="https://opendev.org/openstack/ironic.git"
 IRONIC_IMAGE_REPO="https://github.com/metal3-io/ironic-image.git"
+IRONIC_IMAGE_REPO_REF="${IRONIC_IMAGE_REPO_REF:-"HEAD"}"
```
fmuyassarov, line 17: just a nit, maybe you could rename it to `IRONIC_IMAGE_REPO_COMMIT`?

fmuyassarov (commented):
Aren't we missing 
```
IRONIC_INSPECTOR_REFSPEC
IRONIC_IMAGE_REPO_REF
IRONIC_IMAGE_BRANCH
IRONIC_REFSPEC
```
from the pipeline?

ci/scripts/image_scripts/run_build_ironic.sh
smoshiur1237: Here it was added this var because  the requirement says that we need to add the repo info in metadata file.

ci/scripts/image_scripts/run_build_ironic.sh
smoshiur1237: This is added for the requirement from ccd.

ci/scripts/image_scripts/run_build_ironic.sh
smoshiur1237: Same reason as previous comment^^.

ci/scripts/image_scripts/run_build_ironic.sh
smoshiur1237: I follow the same format in JJB file, `IPA_REPO_REF` as commit hash var. If you guys agree, can change. @Rozzii

ci/scripts/image_scripts/run_build_ironic.sh
fmuyassarov: correct, but I don't see any benefit of adding it. But not blocking too, if you think it is needed let's keep it

Rozzii: I am not sure how clear is it for outsiders that where do we get the source code from so that is why i think it is a good idea to put the inspector repo adress into the metadata.

fmuyassarov: +1 for that case

smoshiur1237: Done

ci/scripts/image_scripts/start_centos_ipa_ironic_build.sh
```diff
@@ -55,7 +55,12 @@ ssh \
   -i "${AIRSHIP_CI_USER_KEY}" \
   "${AIRSHIP_CI_USER}"@"${BUILDER_IP}" \
   PATH=/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/sbin:/bin \
-  IRONIC_REFSPEC="${IRONIC_REFSPEC:-}" DOCKER_USER="${DOCKER_USER}" \
+  IRONIC_REFSPEC="${IRONIC_REFSPEC:-}" \
+  IRONIC_IMAGE_REPO="${IRONIC_IMAGE_REPO:-}" \
```
smoshiur1237, line 59: Is it necessary to pass `IRONIC_IMAGE_REPO`  here. As this var will get the value from `run_ibuild_ironic.sh` file.

Rozzii (approved):

fmuyassarov (approved):

PR #407, opened 2021-10-04
Add openstack node image building in Frankfurt

kashifest:
This PR adds openstack node image building in Frankfurt.


macaptain (approved):

kashifest:
This is actually not needed since the image is already built and kept in artifactory. Closing this PR

PR #408, opened 2021-10-04
Fix issues related to ipa ipeline improvements

Rozzii:
This change adds the following fixes:
  - Fixes missing metal3-dev-env repository commit variable
  - Fixes missing ironic-inspector repository commit variable
  - Adds default refspecs for ironic related repositories
  - Organizes the IPA '.tar' archive's contet to a single directory
    inside the archive.
  - Fix shellcheck warnings


namnx228 (commented):
LGTM

ci/scripts/image_scripts/run_build_ironic.sh
```diff
@@ -19,12 +19,20 @@ IRONIC_IMAGE_BRANCH="${IRONIC_IMAGE_BRANCH:-"master"}"
...
-docker login "${IMAGE_REGISTRY}" -u $DOCKER_USER -p $DOCKER_PASSWORD
+docker login "${IMAGE_REGISTRY}" -u "${DOCKER_USER}" -p "${DOCKER_PASSWORD}"
+
+# Gather ironic-inspector commit hash
+git clone "${IRONIC_INSPECTOR_REPO}"
+pushd "ironic-inspector"
+git fetch origin "${IRONIC_INSPECTOR_REFSPEC}"
+git checkout FETCH_HEAD
+IRONIC_INSPECTOR_COMMIT="$(git rev-parse HEAD)"
+popd
```
fmuyassarov, line 30: I think IRONIC_INSPECTOR_REFSPEC would already be enough info about the state, but since we are doing the same for Ironic repo too, I'm okay with it.

ci/scripts/image_scripts/build_ipa.sh
```diff
@@ -142,8 +145,8 @@ EOF
     popd
     pushd "${IPAMPATH}"
     cat << EOF >> "${METADATA_PATH}"
-IPAM_REPO="${IPAMREPO}"
-IPAM_BRANCH="${IPAMBRANCH}"
+IPAM_REPO="${IPAMREPO:-}"
+IPAM_BRANCH="${IPAMBRANCH:-}"
```
fmuyassarov, line 149: what about capm3, and bmo?

ci/scripts/image_scripts/run_build_ironic.sh
Rozzii: It would be enough most of the time except when the default refspec is used because that points to master and it would be unknown that what was the actual commit under HEAD.

ci/scripts/image_scripts/build_ipa.sh
Rozzii: It was a shellcheck issue as shellcheck was throwing warnings that it might be a type of IPA_REPO and IPA_BRANCH so it is here to make parsing simpler for shellcheck.

ci/scripts/image_scripts/run_build_ironic.sh
```diff
@@ -9,8 +9,8 @@ sudo apt update -y
 # Example: refs/changes/74/804074/1
 IMAGE_REGISTRY="registry.nordix.org"
 CONTAINER_IMAGE_REPO="airship"
-IRONIC_REFSPEC="${IRONIC_REFSPEC:-}"
-IRONIC_INSPECTOR_REFSPEC="${IRONIC_INSPECTOR_REFSPEC:-}"
+IRONIC_REFSPEC="${IRONIC_REFSPEC:-+refs/heads/master:refs/remotes/origin/master}"
+IRONIC_INSPECTOR_REFSPEC="${IRONIC_INSPECTOR_REFSPEC:-+refs/heads/master:refs/remotes/origin/master}"
```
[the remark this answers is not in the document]
Rozzii: Thanks, you are right it defaults to fetch the master, I just added this because if it is left empty then these refspec will show up as empty strings in metadata when the default refspec is fetched.

ci/scripts/image_scripts/run_build_ironic.sh
fmuyassarov: Makes sense.

ci/scripts/image_scripts/build_ipa.sh
fmuyassarov: Got it. I was just asking that would not the same issue/warning occur for CAPM3_REPO, CAPM3_BRANCH vars too

Rozzii: Interestingly no, for some reason it has problem parsing only these two

fmuyassarov (approved):

ci/scripts/image_scripts/build_ipa.sh
Rozzii: I think I will add the ':-' because it might be confusing for others in the future.

Rozzii:
Added the ':-' to variables that are sourced from dev-env to make the code more  unified. I have also fixed a typo in the commit message.

ci/scripts/image_scripts/build_ipa.sh
```diff
@@ -154,7 +157,8 @@ EOF
     popd
 fi
 
-tar -rf "${IPA_IMAGE_TAR}" "${METADATA_PATH}"
+mv "${METADATA_PATH}" "metadata.txt"
```
fmuyassarov, line 160: was there an issue with tarring from `tmp/` dir?

fmuyassarov (approved):

PR #418, opened 2021-10-07
 Add ubuntu user to docker group

mboukhalfa:


ci/images/gen_base_ubuntu_image.sh
```diff
@@ -1,6 +1,6 @@
 #! /usr/bin/env bash
-
 set -eux
+export KUBERNETES_VERSION="v1.22.2"
```
Sunnatillo, line 3: Could you please change the version to v1,22.3?

Sunnatillo (changes requested):

PR #423, opened 2021-10-11
Add hosting container images in Harbor script

furkatgofurov7:
This will be run as part of already existing docker_image_building pipeline we have that is run once every day because this pipeline already does push some container images (lint-md, lint-go) to the Nordix Harbor registry. 
The digest of the container image will be used as tag before pushing since that is unique, compared to having always the same tag in container images (i.e master or latest), using the latter we will not be able to differentiate container images whether they are synced with quay.io or not. 

Update:
The real motivation behind this patch is, certain images that we are using are hosted in quay.io, and sometimes we see service outage in the quay.io registry. For that reason, we want to host identical copies of all container images in Nordix Harbor registry to be used as a standby in case of quay.io inaccessibility.

### DIFF
--- a/ci/scripts/image_scripts/synchronize_container_images.sh
+++ b/ci/scripts/image_scripts/synchronize_container_images.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+set -eu
+
+# This script makes sure to synchronize container images hosted in Nordix/Harbor
+# with the ones hosted in quay.io/metal3 based on the container image digest as tag.
+
+# Container image related variables
+UPSTREAM_IMAGE_REGISTRY="quay.io"
+UPSTREAM_CONTAINER_IMAGE_REPO="metal3-io"
+HARBOR_IMAGE_REGISTRY="registry.nordix.org"
+HARBOR_CONTAINER_IMAGE_REPO="airship"
+CAPM3_v1a4_RELEASE_TAG="release-0.4"
+IPAM_v1a4_RELEASE_TAG="release-0.0"
+CAPM3_IPAM_v1a5_RELEASE_TAG="master"
+
+# Container runtime
+export CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-docker}"
+
+# Declare a string array with container image names
+declare -a ContainerImagesArray=(
+    "vbmc" "sushy-tools" "ironic-ipa-downloader" "ironic" \
+    "ironic-client" "keepalived" "baremetal-operator" \
+    "cluster-api-provider-metal3:${CAPM3_v1a4_RELEASE_TAG}" \
+    "cluster-api-provider-metal3:${CAPM3_IPAM_v1a5_RELEASE_TAG}" \
+    "ip-address-manager:${IPAM_v1a4_RELEASE_TAG}" \
+    "ip-address-manager:${CAPM3_IPAM_v1a5_RELEASE_TAG}"
+)
+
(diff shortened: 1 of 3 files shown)


macaptain (dismissed):

namnx228:
@furkatgofurov7 The patch looks good to me. Also, could you please add some explanations about the motivation of pulling the images from quay.io then pushing them to Nordix Harbor?

ci/scripts/image_scripts/synchronize_container_images.sh
```diff
@@ -0,0 +1,49 @@
...
+export CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-docker}"
+
+# Declare a string array with container image names
+declare -a ContainerImagesArray=(
+    "vbmc" "sushy-tools" "ironic-ipa-downloader" "ironic" \
+    "ironic-client" "keepalived" "baremetal-operator" \
+    "cluster-api-provider-metal3:${CAPM3_v1a4_RELEASE_TAG}" \
+    "cluster-api-provider-metal3:${CAPM3_IPAM_v1a5_RELEASE_TAG}" \
+    "ip-address-manager:${IPAM_v1a4_RELEASE_TAG}" \
+    "ip-address-manager:${CAPM3_IPAM_v1a5_RELEASE_TAG}"
```
kashifest, line 27: Can we rename the var to `v1a5_RELEASE_TAG` only? Since it is master. Cosmetic change

ci/scripts/image_scripts/synchronize_container_images.sh
```diff
@@ -0,0 +1,49 @@
...
+    # Pull container image from quay.io
+    "${CONTAINER_RUNTIME}" pull "${UPSTREAM_IMAGE_REGISTRY}/${UPSTREAM_CONTAINER_IMAGE_REPO}/${container}"
+    # Get the sha256 digest of the container image
+    digest=$("${CONTAINER_RUNTIME}" inspect --format='{{index .RepoDigests 0}}' \
+        "${UPSTREAM_IMAGE_REGISTRY}/${UPSTREAM_CONTAINER_IMAGE_REPO}/${container}" \
+        | grep -o 'sha256:.*' | cut -f2- -d: | cut -c1-6)
+    echo "$container container image digest is $digest"
+    # Tag and push quay container image only if it doesn't already exist in the Nordix Harbor registry.
+    if "${CONTAINER_RUNTIME}" manifest inspect "${HARBOR_CONTAINER_IMAGE_REPO}/$container:${digest}"  > /dev/null; then
+        echo "${HARBOR_CONTAINER_IMAGE_REPO}/$container:${digest} container image already exists -> skip pushing container image to Nordix Harbor registry"
```
kashifest, line 41: Do we need this check? isn't is so that its no-op if it exists?

kashifest (commented):
lgtm. Minor comments

ci/scripts/image_scripts/synchronize_container_images.sh
furkatgofurov7: Yes, we do to prevent pushing same image again, so we will not push image to registry but just output the message to jenkins log

ci/scripts/image_scripts/synchronize_container_images.sh
furkatgofurov7: Done

smoshiur1237 (approved):
Good improvement and LGTM.

PR #424, opened 2021-10-12
Refactor OpenStack volume creation

macaptain:
It should be possible to create a volume from an image in OpenStack without booting any machines.

Every time a VM is booted in OpenStack from an image, the image is copied to a volume. The volume is attached to a VM which is booted. We do volume creation in the image build so that the the image to volume copying step does not need to be repeated for every instance boot.

Since we no longer spin up a VM to do the volume creation step, the cleanup_volume.sh script is no longer needed. (The packer image creation does cleanup of VMs for the image building part of the pipeline on image creation failure.)

I've kept the retry behavior from the old script, since this offers protection from any CityCloud flakes.


ci/images/gen_metal3_centos_volume.sh
```diff
@@ -2,37 +2,18 @@
 
 set -eu
 
-SSH_PRIVATE_KEY_FILE="${1:?}"
-USE_FLOATING_IP="${2:?}"
-
-CI_DIR="$(dirname "$(readlink -f "${0}")")/.."
-IMAGES_DIR="${CI_DIR}/images"
-OS_SCRIPTS_DIR="${CI_DIR}/scripts/openstack"
```
furkatgofurov7, line 10: This is used below

ci/images/gen_metal3_ubuntu_volume.sh
```diff
@@ -2,37 +2,18 @@
 
 set -eu
 
-SSH_PRIVATE_KEY_FILE="${1:?}"
-USE_FLOATING_IP="${2:?}"
-
-CI_DIR="$(dirname "$(readlink -f "${0}")")/.."
-IMAGES_DIR="${CI_DIR}/images"
-OS_SCRIPTS_DIR="${CI_DIR}/scripts/openstack"
```
furkatgofurov7, line 10: same

macaptain:
> Maybe I am not getting the idea here, but what are you trying to achieve here, to me it is only creating a bootable volume? Where do we run 01_* to decrease the timing, which was the sole main motivation behind using them.

It's run here: https://github.com/Nordix/airship-dev-tools/blob/ebd19dbe486ddac6e07055c63f1cdab77feb5467/ci/scripts/image_scripts/provision_metal3_image.sh#L53
And here: https://github.com/Nordix/airship-dev-tools/blob/ebd19dbe486ddac6e07055c63f1cdab77feb5467/ci/scripts/image_scripts/provision_metal3_image_centos.sh#L57

This provision script is run by packer (see `REMOTE_EXEC_CMD`). The output of the image build already has 01 run. We only need to convert that image to a volume.

kashifest:
> Maybe I am not getting the idea here, but what are you trying to achieve here, to me it is only creating a bootable volume? Where do we run 01_* to decrease the timing, which was the sole main motivation behind using them.

Its now done as part of the image building process. It landed with this #397

furkatgofurov7 (approved):
/approve

PR #425, opened 2021-10-12
Remove key env var in volume building steps

macaptain:


kashifest (commented):
Can you elaborate a bit in the PR description , why we dont need this?

PR #429, opened 2021-10-21
Increase image build timeouts to 90 minutes

macaptain:
Previously, the image build timeout was set to 60 minutes. However, this isn't enough for the CentOS image build in the Kna1 region to reliably finish, so let's bump it to 90 minutes.


kashifest:
This has been enough previously. Not sure if bumping this is a good idea or would help the cause.

macaptain:
> This has been enough previously. Not sure if bumping this is a good idea or would help the cause.

After discussing, hopefully this change will improve reliability without many drawbacks.

PR #441, opened 2021-11-16
Add syncing for capm3 release branches

Rozzii:
This commit also makes the sync script shellcheck compatible.

### DIFF
--- a/scripts/update-nordix-repos-master.sh
+++ b/scripts/update-nordix-repos-master.sh
@@ -8,7 +8,7 @@ set -u
 # -push jenkins local repos to Nordix forks
 #------------------------------------------
 
-WORKSPACE=${WORKSPACE:=/tmp}
+WORKSPACE="${WORKSPACE:=/tmp}"
 
 SCRIPTPATH="$(dirname "$(readlink -f "${0}")")"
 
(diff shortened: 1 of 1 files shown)


scripts/update-nordix-repos-master.sh
```diff
@@ -115,4 +117,20 @@ done
...
+CAPM3_RELEASE_BRANCHES="release-0.3 release-0.4 release-0.5"
+i=0
+for branch in ${CAPM3_RELEASE_BRANCHES}
+do
+  echo "Updating ${branch} branch in ${LOCAL_CAPM3_REPO}"
+  pushd "${LOCAL_CAPM3_REPO}" || exit
+  git checkout "origin/${branch}"
+  git fetch "origin" "${branch}"
+  git rebase "origin" "${branch}"
+  git remote add "nordixrepo${i}" "${NORDIX_CAPM3_REPO}"
```
fmuyassarov, line 129: could have been also```suggestion
  git remote add "nordixrepo${branch}" "${NORDIX_CAPM3_REPO}"
```
, but I'm fine with the current approach too

**Review by fmuyassarov — APPROVED** [review]
It seems that we are missing release-0.5 branch in downstream fork. Shall we create that first before kicking of the job?

File: scripts/update-nordix-repos-master.sh [hunk 2]
```diff
@@ -115,4 +117,20 @@ done
...
+  git push "nordixrepo${i}" "HEAD:${branch}"
+  echo "Push done to ${branch} branch in ${NORDIX_CAPM3_REPO}"
+  popd || exit
+  i=$((i+1))
+done
```
furkatgofurov7, line 134: Nice, we could do the same synchronizing for IPAM repo branches (release-0.0, release-0.1)

furkatgofurov7: In the same or separate patch of course. If you do in this patch, please update patch title.

scripts/update-nordix-repos-master.sh
Rozzii: I just wanted to avoid using special chars in the remote name e.g. " - " and  ".".

scripts/update-nordix-repos-master.sh
Rozzii: Sure I can do it in a followup PR, I just want to test that this works first.

furkatgofurov7 (approved):
/approve

scripts/update-nordix-repos-master.sh
Rozzii: But now I have tested it can handle special chars so I have changed it according to your suggestion.

fmuyassarov: thanks

fmuyassarov (approved):

PR #480, opened 2022-02-07
Add metal3ci user to image userdata

macaptain:
This adds a metal3ci user to image userdata. The airshipci user will still be created and used by the scripts. This additional user will let us swap over to using the metal3ci user once images have been built with both users.

### DIFF
--- a/ci/images/centos_userdata.tpl
+++ b/ci/images/centos_userdata.tpl
@@ -6,6 +6,12 @@ users:
     sudo: ['ALL=(ALL) NOPASSWD:ALL']
     groups: ${DEFAULT_SSH_USER_GROUP}
     shell: /bin/bash
+  - name: metal3ci
+    ssh-authorized-keys:
+      - ${SSH_AUTHORIZED_KEY}
+    sudo: ['ALL=(ALL) NOPASSWD:ALL']
+    groups: wheel
+    shell: /bin/bash
 
 runcmd:
   - sed -i "/^127.0.0.1/ s/$/ ${HOSTNAME}/" /etc/hosts
--- a/ci/images/ubuntu_userdata.tpl
+++ b/ci/images/ubuntu_userdata.tpl
@@ -6,6 +6,12 @@ users:
     sudo: ['ALL=(ALL) NOPASSWD:ALL']
     groups: ${DEFAULT_SSH_USER_GROUP}
     shell: /bin/bash
+  - name: metal3ci
+    ssh-authorized-keys:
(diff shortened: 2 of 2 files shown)


namnx228 (approved):

ci/images/ubuntu_userdata.tpl
```diff
@@ -6,6 +6,12 @@ users:
     sudo: ['ALL=(ALL) NOPASSWD:ALL']
     groups: ${DEFAULT_SSH_USER_GROUP}
     shell: /bin/bash
+  - name: metal3ci
+    ssh-authorized-keys:
+      - ${SSH_AUTHORIZED_KEY}
+    sudo: ['ALL=(ALL) NOPASSWD:ALL']
+    groups: sudo
```
furkatgofurov7, line 13: user groups varying between ubuntu and centos, is that on purpose?

macaptain: Yes, the root group in Ubuntu is `sudo`, but in CentOS, it's `wheel`:
https://github.com/Nordix/metal3-dev-tools/blob/2a6ed3e119f0c269fdce1085cbd5a46868bdf86a/ci/images/gen_metal3_ubuntu_image.sh#L28
https://github.com/Nordix/metal3-dev-tools/blob/2a6ed3e119f0c269fdce1085cbd5a46868bdf86a/ci/images/gen_metal3_centos_image.sh#L48

ci/images/centos_userdata.tpl
```diff
@@ -6,6 +6,12 @@ users:
     sudo: ['ALL=(ALL) NOPASSWD:ALL']
     groups: ${DEFAULT_SSH_USER_GROUP}
     shell: /bin/bash
+  - name: metal3ci
```
kashifest, line 9: why hardcoding this?

ci/images/ubuntu_userdata.tpl
furkatgofurov7: We could use those parameterized vars in that case

macaptain: What's the advantage of using a var instead of a string?

kashifest (approved):

macaptain:
There's a user variable `CI_SSH_USER_NAME` which will eventually be switched to 'metal3ci' instead of 'airshipci', I will update that variable after we've swapped over fully to that user, so this user will be parameterized in the template later to be consistent across the scripts, but it's a 2 step process. Will merge this and check the image build is all good.

PR #492, opened 2022-02-18
Replace TARGET_NODE_OS with IMAGE_OS

Sunnatillo:
Replace TARGET_NODE_OS with IMAGE_OS since TARGET_NODE_OS variable removed from jjb


lentzi90 (approved):
LGTM

PR #498, opened 2022-02-28
fix incorrect metadata concatenation

Rozzii:


smoshiur1237 (dismissed):

lentzi90:
Why is it wrong to overwrite the file if we know this is all the data we want to have in it?

Rozzii:
> Why is it wrong to overwrite the file if we know this is all the data we want to have in it?

It is not the first place where we write the file but actually it could be I can combine the two write actions!

lentzi90 (approved):

PR #510, opened 2022-03-23
Update CAPO docs and scripts

lentzi90:
- Bumped CAPI version
- Bumped kubernetes version
- Bumped yq version (v4 is now supported in CAPO)
- Updated docs and added cleanup instructions

### DIFF
--- a/capo/README.md
+++ b/capo/README.md
@@ -4,10 +4,11 @@
 
 * Install go version >= 1.16
 * Install `kind` and `tilt`
-* Install version 3.x of `yq` (note: version 4.x is not compatible)
+* Install `yq`
 
-```
-wget https://github.com/mikefarah/yq/releases/download/3.4.1/yq_linux_amd64
+```bash
+YQ_VERSION="4.23.1"
+wget "https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/yq_linux_amd64"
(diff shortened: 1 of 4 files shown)


capo/README.md
```diff
@@ -110,6 +105,27 @@ openstack extension list -f json -c Description -f value | grep trunk
   Expose trunk port details
 ```
 
+## Cleanup
+
+The "normal" way to clean up openstack resources created by CAPO is to delete all Cluster objects you created.
+The default namespace is `default` and the default cluster name is `basic-1`.
+If you followed the suggestion to change the name of the cluster it would be whatever you chose, e.g. `${USER}-capo-test`.
```
smoshiur1237, line 112: same "you"

smoshiur1237 (changes requested):

lentzi90:
Thanks for the review @smoshiur1237 ! I'm a bit hesitant to remove the "you" here because we already had it in multiple places before my changes.
These docs are very specific for developers in our team, not general user guides, so I think it is ok to refer to the developer as "you". It makes it clearer what the developer can choose (e.g. cluster name), and what is common guidelines that everyone has to follow.

smoshiur1237 (commented):
LGTM

smoshiur1237 (approved):

PR #512, opened 2022-03-29
Use distro-sync instead of update on centos

lentzi90:
Also switched yum to dnf (we already used dnf in one place and yum is
linked to dnf).

We got the following error when building the centos image:

```
Error:
 Problem: package centos-stream-repos-8-4.el8.noarch requires centos-gpg-keys = 1:8-4.el8, but none of the providers can be installed
  - cannot install both centos-gpg-keys-1:8-5.el8.noarch and centos-gpg-keys-1:8-4.el8.noarch
  - cannot install the best update candidate for package centos-stream-repos-8-2.el8.noarch
  - cannot install the best update candidate for package centos-gpg-keys-1:8-2.el8.noarch
```

According to [this](https://linux-community.web.cern.ch/t/centos8-dnf-update-fails-any-suggestion/120), using `distro-sync` should solve the dependency issue.


furkatgofurov7 (approved):
Can you also add the image build failure log in the PR description? Thanks!

/approve

PR #514, opened 2022-04-13
Update script for building metal3 images locally

lentzi90:
- Use source image name directly instead of searching for the ID based on the name.
- Remove hard coded source image
- Parameterize ssh_keypair_name and reuse_ips
- Provide overridable defaults for local script
- Don't upload locally built images to artifactory

### DIFF
--- a/ci/images/README.md
+++ b/ci/images/README.md
@@ -84,3 +84,40 @@ The centos building scripts take three arguments :
    ./gen_<xxx>_<centos>_image.sh /data/keys/id_rsa_metal3ci 1 <provisioner script>
    ```
 
+### Building and testing locally
+
+The scripts mentioned above (gen_*_image.sh) are made for use in our CI pipelines.
+As such, they make certain assumptions that may not be ideal when developing and testing things locally (e.g. keypair name and ssh-key path).
+
+For these situations there is another script: `run_local.sh`.
+It allows overriding many variables and skips things like uploading the images to Artifactory.
+
+This is how you use it:
+
+1. Check the comments and variables at the top of `run_local.sh` and determine what you want/need to override.
+2. Create a file with your custom variables.
+3. Get an openstack.rc file with credentials to the cloud you want to build in.
+4. Source your variables, source the openstack.rc.
+5. Run the script: `./run_local.sh <provisioning-script>`
+
+Here is an example of custom variables:
+
+```shell
+# My custom variables
+export SSH_USER_NAME="foo"
+export SSH_USER_GROUP="wheel"
+export SSH_KEYPAIR_NAME="foo-key"
+export SSH_PUBLIC_KEY_FILE="/home/foo/.ssh/id_rsa.pub"
+export SSH_PRIVATE_KEY_FILE="/home/foo/.ssh/id_rsa"
+export IMAGE_NAME="foo-test-image"
+export SOURCE_IMAGE_NAME="Upstream-centos-or-ubuntu"
+export NETWORK="abcdefg"
+```
+
+And here is how to use it:
+
+```console
+$ source vars.sh
+$ source openstack.rc
+$ ./run_local.sh provision_metal3_image.sh
+```
--- a/ci/images/gen_base_ubuntu_image.sh
+++ b/ci/images/gen_base_ubuntu_image.sh
@@ -28,7 +28,6 @@ FLOATING_IP_NETWORK="$( [ "${USE_FLOATING_IP}" = 1 ] && echo "${EXT_NET}")"
 REMOTE_EXEC_CMD="KUBERNETES_VERSION=${KUBERNETES_VERSION} /home/${SSH_USER_NAME}/image_scripts/provision_base_image.sh"
 SSH_USER_GROUP="sudo"
 
-SOURCE_IMAGE="$(get_resource_id_from_name image "${SOURCE_IMAGE_NAME}")"
 SSH_AUTHORIZED_KEY="$(cat "${OS_SCRIPTS_DIR}/id_rsa_metal3ci.pub")"
 render_user_data \
   "${SSH_AUTHORIZED_KEY}" \
@@ -50,7 +49,7 @@ create_keypair "${CI_PUBLIC_KEY_FILE}" "${SSH_KEYPAIR_NAME}"
 # Build Image
 packer build \
   -var "image_name=${IMAGE_NAME}" \
-  -var "source_image=${SOURCE_IMAGE}" \
+  -var "source_image_name=${SOURCE_IMAGE_NAME}" \
   -var "user_data_file=${USER_DATA_FILE}" \
   -var "exec_script_path=${STARTER_SCRIPT_PATH}" \
   -var "ssh_username=${SSH_USER_NAME}" \
--- a/ci/images/image_builder_template.json
+++ b/ci/images/image_builder_template.json
@@ -1,7 +1,7 @@
 {
   "variables": {
     "image_name": "",
-    "source_image": "4654e14c-f667-4e24-a538-c792b57c5bc8",
+    "source_image_name": "",
     "user_data_file": "userdata",
     "exec_script_path": "",
     "ssh_username": "metal3ci",
@@ -9,6 +9,7 @@
     "ssh_private_key_file": "",
     "network": "282a8c18-614c-4fbc-9928-b66d56ae3cbe",
     "floating_ip_net": "",
+    "reuse_ips": "false",
     "local_scripts_dir": "../scripts",
     "ssh_pty": "false",
     "flavor":"2C-16GB-20GB"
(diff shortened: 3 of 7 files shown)


ci/images/gen_base_ubuntu_image.sh
```diff
@@ -28,7 +28,6 @@ FLOATING_IP_NETWORK="$( [ "${USE_FLOATING_IP}" = 1 ] && echo "${EXT_NET}")"
 REMOTE_EXEC_CMD="KUBERNETES_VERSION=${KUBERNETES_VERSION} /home/${SSH_USER_NAME}/image_scripts/provision_base_image.sh"
 SSH_USER_GROUP="sudo"
 
-SOURCE_IMAGE="$(get_resource_id_from_name image "${SOURCE_IMAGE_NAME}")"
```
furkatgofurov7, line 31: This was a useful utility I believe and used for networks too, could you explain a bit more what's the motivation/benefit of this change?

lentzi90: The utility is still there and used for networks. For source image, it is not needed however, since we can set `source_image_name` instead of `source_image` on line 52/53 below.

ci/images/image_builder_template.json
```diff
@@ -1,26 +1,27 @@
 {
   "variables": {
     "image_name": "",
-    "source_image": "4654e14c-f667-4e24-a538-c792b57c5bc8",
+    "source_image_name": "",
```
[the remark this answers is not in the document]
lentzi90: I think it is not very useful to set a value here. In the CI we always override it so there is no need for it there. When building locally it is almost guaranteed that you want to change this, since the image names are different between cloud providers and regions. The old ID that I removed points to an image from 2019 with ubuntu 18.04 that I don't think we have been using for a long time. So my point is that it's probably better to leave it empty so that the user will notice and set a proper value.

ci/images/run_local.sh
```diff
@@ -10,27 +10,73 @@
 
 # For example:
 # $ source openstack.rc
-# $ ./run_local.sh gen_metal3_centos_volume.sh ~/keys/metal3ci_id_rsa
+# $ ./run_local.sh provision_metal3_image_centos.sh
 
 set -eux
 
-SCRIPT="${1}"
-KEY_PATH="${2}"
```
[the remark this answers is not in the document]
lentzi90: The default below is just the default, it can be easily overridden by setting `SSH_KEYPAIR_NAME`. The reason I removed it from the arguments is that I think it makes more sense to treat it the same way as the rest of the variables. But maybe it would be better to not set a default at all and force the user to set them, what do you think?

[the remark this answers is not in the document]
furkatgofurov7: I think that would make life easier for someone trying this out with non-EST credentials set or for a single user with multiple SSH keys

ci/images/gen_base_ubuntu_image.sh
furkatgofurov7: That is clear, point was why did we need that utility at all if we could simple set it directly. Checking more into it, I found out it returns image/network ID and we pass it later to packer. Going further, maybe it is better to play with unique IDs' rather than names which is not guaranteed to be unique?

ci/images/image_builder_template.json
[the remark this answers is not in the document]
furkatgofurov7: Got it, should we mention it may be at the beginning of the _run_local_ script in a comments section that someone trying to build the local image, should pass it themselves?

ci/images/gen_base_ubuntu_image.sh
lentzi90: I don't understand. Previously we set `SOURCE_IMAGE_NAME`, then we run this custom function to get the ID from the name and put it in `SOURCE_IMAGE` that we pass to packer. Now we set `SOURCE_IMAGE_NAME` and pass it directly to packer. I don't see how there could be a difference since we start with the image name in both cases. Either packer, openstack or our custom function will find the ID from the name. :thinking: 

If we want to be absolutely sure what image is used, we should of course use ID everywhere, but this was not the case before either.

furkatgofurov7: > If we want to be absolutely sure what image is used, we should of course use ID everywhere, but this was not the case before either.

Why not? This was the exact case:
```
  local RESOURCE_TYPE RESOURCE_NAME RESOURCE_INFO

  RESOURCE_TYPE="${1:?}"
  RESOURCE_NAME="${2:?}"

  # Get Rsource Info
  # shellcheck disable=SC2086
  RESOURCE_INFO="$(openstack ${RESOURCE_TYPE} show "${RESOURCE_NAME}" -f json)"

  [[ -z "${RESOURCE_INFO}" ]] && return

  # Output list of IDs
  echo "${RESOURCE_INFO}" | jq -r '.id'
}
```

we fetched the ID and set it to `SOURCE_IMAGE` which is passed to packer later

lentzi90: My point is that we got the ID based on the name, so it is no better than using the name directly. If there are multiple resources with the same name, this function cannot know which is the correct one.

furkatgofurov7: Is packer smart enough to find out the ID itself when just the name is passed?

lentzi90: I'm not sure if it is packer or openstack, I would actually assume that it is openstack since it is possible to pass either name or ID to `openstack server create`. But yes it is possible to only give the name to packer and it will work. This is what I'm changing here to avoid the extra function call.

ci/images/run_local.sh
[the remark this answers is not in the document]
lentzi90: Updated so that SSH_KEYPAIR_NAME is required and added more comments to explain that some variables are expected to be set and some can be overridden. PTAL :slightly_smiling_face:

ci/images/run_local.sh
```diff
@@ -4,33 +4,89 @@
...
-METAL3_CI_USER="metal3ci"
-RT_URL="https://artifactory.nordix.org/artifactory"
-OS_AUTH_URL="https://kna1.citycloud.com:5000"
-OS_USER_DOMAIN_NAME="CCP_Domain_37137"
-OS_PROJECT_DOMAIN_NAME="CCP_Domain_37137"
-OS_REGION_NAME="Kna1"
-OS_PROJECT_NAME="Default Project 37137"
-OS_TENANT_NAME="Default Project 37137"
-OS_AUTH_VERSION=3
-OS_IDENTITY_API_VERSION=3
```
furkatgofurov7, line 30: One last question, where are we sourcing these from if removed?

furkatgofurov7:
> * Don't upload locally built images to artifactory

Why not, most of the time the purpose of building it locally is to try it out in CI and see how it works?

ci/images/run_local.sh
lentzi90: There is a comment at the top that the user should source openstack.rc (their own openstack credentials) before running the script. This comment was even there before so I'm not sure why these variables were here. They would override everything except username/password.

lentzi90:
> > * Don't upload locally built images to artifactory
> 
> Why not, most of the time the purpose of building it locally is to try it out in CI and see how it works?

Good question. I realize that I may not have thought about all use cases. I was using this script to build the metal3 images, but not node images and I guess there is the difference. We only need the metal3 image in openstack in order to test them so I saw no point in uploading to artifactory.

For node images I guess we need to upload them to artifactory in order to try them. But we do have a pipeline that does this automatically already (`test-integration-metal3-dev-tools-{image_os}`). And the user can still manually upload it using `upload_node_image_rt.sh` if necessary.

After thinking through this I would prefer actually to avoid uploading it automatically. We avoid the need to artifactory credentials and there is no risk forgetting to clean up old images. But I should probably add something about this local building in the readme...

furkatgofurov7:
> For node images I guess we need to upload them to artifactory in order to try them. But we do have a pipeline that does this automatically already (`test-integration-metal3-dev-tools-{image_os}`). And the user can still manually upload it using `upload_node_image_rt.sh` if necessary.

Good point, totally forgot about that pipeline, that actually simplifies local node image building and testing a lot (building with custom params, uploading it, running integration tests using built node images and delete them at the end of the run) where we can run it on a CI. 

> But I should probably add something about this local building in the readme...

+1

ci/images/run_local.sh
furkatgofurov7: These were the defaults most probably in case .rc file misses any of them

lentzi90: Well maybe that was the intent, but it was actually overriding all of it. :grimacing:

lentzi90:
Thanks for the reviews @furkatgofurov7 ! I have now added a section to the readme to explain how to use the script :slightly_smiling_face:

ci/images/README.md
```diff
@@ -84,3 +84,40 @@ The centos building scripts take three arguments :
...
+
+This is how you use it:
+
+1. Check the comments and variables at the top of `run_local.sh` and determine what you want/need to override.
+2. Create a file with your custom variables.
+3. Get an openstack.rc file with credentials to the cloud you want to build in.
+4. Source your variables, source the openstack.rc.
+5. Run the script: `./run_local.sh <provisioning-script>`
+
+Here in a example for custom variables:
```
furkatgofurov7, line 103:
```suggestion
Here is an example of custom variables:
```

furkatgofurov7 (commented):
Thanks @lentzi90 last minor suggestions from my side:

furkatgofurov7 (approved):

PR #515, opened 2022-04-13
Enable colored output for pipelines that use color

lentzi90:
Integration tests and image building pipelines are already using colored output, which makes the logs hard to read without this.

### DIFF
--- a/ci/jobs/metal3_dev_tools_integration_tests.pipeline
+++ b/ci/jobs/metal3_dev_tools_integration_tests.pipeline
@@ -20,6 +20,7 @@ script {
 
 pipeline {
   agent { label 'metal3-static-workers' }
+  options { ansiColor('xterm') }
   environment {
     METAL3_CI_USER="metal3ci"
     VM_KEY = "${VM_KEY}"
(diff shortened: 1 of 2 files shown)


ci/jobs/metal3_dev_tools_integration_tests.pipeline
```diff
@@ -165,15 +166,13 @@ pipeline {
         withCredentials([usernamePassword(credentialsId: 'metal3ci_city_cloud_openstack_credentials', usernameVariable: 'OS_USERNAME', passwordVariable: 'OS_PASSWORD')]) {
           withCredentials([sshUserPrivateKey(credentialsId: 'metal3ci_city_cloud_ssh_keypair', keyFileVariable: 'METAL3_CI_USER_KEY')]) {
             withCredentials([string(credentialsId: 'metal3-clusterctl-github-token', variable: 'GITHUB_TOKEN')]) {
-              ansiColor('xterm') {
```
lentzi90, line 168: Moved to the top level since other steps also need this.

ci/jobs/openstack_image_building.pipeline
```diff
@@ -10,6 +10,7 @@ script {
 
 pipeline {
   agent { label 'metal3-static-workers' }
+  options { ansiColor('xterm') }
```
furkatgofurov7, line 13: How beneficial is it to enable this on other pipelines(node/ipa building image) as well?

Sunnatillo (approved):
LGTM

furkatgofurov7 (approved):
/approve

ci/jobs/openstack_image_building.pipeline
lentzi90: I think it doesn't do much if the pipeline has no color codes. Haven't seen any issues with it. Do you think it would be better to only set it for specific parts that use color?

furkatgofurov7: Have not checked if we have color codes on those, may be fine to skip it in that case.

PR #516, opened 2022-04-20
Fix cloud-init error on centos

lentzi90:
resolv.conf was previously made immutable to prevent NetworkManager
and/or cloud-init from modifying it. This was causing issues with
cloud-init, since it still tried to modify it.

Most of the time things worked ok (except that cloud-init had error
status). But every now and then this error seemed to cause other issues.
E.g. the resizing of the filesystem or injection of authorized
ssh keys failed.

This commit properly configures NetworkManager to set the nameservers
and cloud-init to avoid touching resolv.conf.

### DIFF
--- a/ci/scripts/image_scripts/provision_metal3_image_centos.sh
+++ b/ci/scripts/image_scripts/provision_metal3_image_centos.sh
@@ -17,6 +17,37 @@ export IMAGE_OS="${IMAGE_OS:-Centos}"
 export EPHEMERAL_CLUSTER="${EPHEMERAL_CLUSTER:-minikube}"
 export CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
 
+# Disable cloud-init network configuration that would overwrite the network config.
+cat <<EOF | sudo tee /etc/cloud/cloud.cfg.d/01-network.cfg
+network:
+  config: disabled
+# resolv.conf is managed by NetworkManager, cloud-init should not touch it
+manage_resolv_conf: false
+EOF
+# Even with the above, cloud-init will configure NetworkManager to not touch
+# resolv.conf *if* there is a nameserver configured for the subnet in openstack.
+# This is done in the following file, that we remove to let NM handle resolv.conf
+# Ref https://bugs.launchpad.net/cloud-init/+bug/1693251
+sudo rm /etc/NetworkManager/conf.d/99-cloud-init.conf || true
+sudo systemctl restart NetworkManager
+# Configure network (set nameservers, disable peer DNS and remove mac address
+# that was automatically added). The rest of the fields are kept as is.
+cat <<EOF | sudo tee /etc/sysconfig/network-scripts/ifcfg-eth0
+NAME="System eth0"
+BOOTPROTO=dhcp
+DEVICE=eth0
+MTU=1500
+ONBOOT=yes
+TYPE=Ethernet
+USERCTL=no
+PEERDNS=no
+DNS1=8.8.8.8
+DNS2=8.8.4.4
+EOF
+# Apply the changes
+sudo nmcli connection reload
+sudo nmcli connection up "System eth0"
+
 sudo dnf distro-sync -y
 sudo dnf update -y curl nss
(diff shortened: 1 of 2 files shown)


lentzi90:
This is to fix the famous "no-space" issue. I have not been able to figure out exactly why an error when trying to configure resolv.conf would cause issues with filesystem resizing unfortunately. My best guess is that there is some kind of interference between them, and depending on how long each task takes it can cause problems for the other.

Anyway, I have an explanation for why this has been happening in the CI only for the Fra1 region. In this region (but in no other) we have configured some nameservers for the subnet in openstack. Because of this, cloud-init will [behave differently](https://bugs.launchpad.net/cloud-init/+bug/1693251) and try to set those nameservers in the resolv.conf, while also configuring NetworkManager to not touch the file. In other regions, where we do not have any nameservers specified, cloud-init will stay out of this and let NetworkManager set things up. In this case the default behavior is to get the nameservers from the router.

The fix is to

1. Configure cloud-init to not touch resolv.conf
2. Configure the proper nameservers using NetworkManager
3. Remove the config that disables NetworkManager DNS management. (This file is currently created by cloud-init during the image building process in Fra1 because of the subnet nameservers mentioned above.)

ci/scripts/image_scripts/provision_metal3_image_centos.sh
```diff
@@ -17,6 +17,27 @@ export IMAGE_OS="${IMAGE_OS:-Centos}"
...
+# Ref https://bugs.launchpad.net/cloud-init/+bug/1693251
+sudo rm /etc/NetworkManager/conf.d/99-cloud-init.conf || true
+sudo systemctl restart NetworkManager
+# Configure nameservers
+sudo nmcli connection modify "System eth0" ipv4.ignore-auto-dns yes
+sudo nmcli connection modify "System eth0" ipv4.dns 8.8.8.8,8.8.4.4
+# Remove mac address since this will be different for each VM
+sudo nmcli c modify "System eth0" 802-3-ethernet.mac-address ""
+# Apply the changes
+sudo nmcli connection up "System eth0"
```
kashifest, line 39: cant we do these configs through some config file rathen than putting it as command?

kashifest:
> But every now and then this error seemed to cause other issues. E.g. the resizing of the filesystem or injection of authorized ssh keys failed.

How do you get that ssh key injection is failing sometimes? I am curious to know since I am facing ssh issue for centos stream 9.

ci/scripts/image_scripts/provision_metal3_image_centos.sh
lentzi90: Yes this should be possible, I'll investigate!

lentzi90:
> > But every now and then this error seemed to cause other issues. E.g. the resizing of the filesystem or injection of authorized ssh keys failed.
> 
> How do you get that ssh key injection is failing sometimes? I am curious to know since I am facing ssh issue for centos stream 9.

This happened to me while testing locally. I have not seen it in the CI, and I think this is because we bake the authorized keys in the image. When testing, I used a different key than the one baked in the image so I needed cloud-init to add the key when creating the VM. In the CI, we don't need it as the key was already baked in the image.

Not sure if that was clear, maybe this diagram will help:

```mermaid
  graph TD;
      A[Generic image]--> |Add authorized key1| B[metal3 image];
      B-->D[Virtual machine];
      C[Cloud-init]-->|Add authorized key2| D;
```

In the CI only key1 is needed. For my local testing I added my personal key as key2. Then I could see that it sometimes failed to add it. I could then still use key1 to get in and debug.

ci/scripts/image_scripts/provision_metal3_image_centos.sh
lentzi90: Fixed!

lentzi90:
/hold
until I verify it is working in https://github.com/metal3-io/project-infra/pull/375

lentzi90:
Working :tada: 
https://jenkins.nordix.org/job/metal3_project_infra_v1b1_integration_test_centos/48/console

PR #522, opened 2022-04-26
Fix nameservers on centos node images

lentzi90:
I totally missed that the node images didn't use the same script when I changed the cloud init. This should fix it.

The reason I didn't add these commands in the user-data, is that cloud-init itself is making network modifications, like disabling NetworkManager DNS. So instead of hoping that it goes in the correct order I think it is better like this.


smoshiur1237 (approved):
/lgtm

PR #532, opened 2022-05-09
Uplift to Ubuntu 22.04

kashifest:


ci/images/ubuntu_userdata.tpl
```diff
@@ -6,12 +6,8 @@ users:
...
-  - name: metal3ci
-    ssh-authorized-keys:
-      - ${SSH_AUTHORIZED_KEY}
-    sudo: ['ALL=(ALL) NOPASSWD:ALL']
-    groups: sudo
-    shell: /bin/bash
 
 runcmd:
   - sed -i "/^127.0.0.1/ s/$/ ${HOSTNAME}/" /etc/hosts
+  - echo "PubkeyAcceptedKeyTypes=+ssh-rsa" >> /etc/ssh/sshd_config
```
elfosardo, line 12: this is ok as a temp fix, but ideally we should switch to a different algorithm as ssh-rsa is deprecated since a while now

kashifest: Yes, I am planning to get rid of this, we need to generate new keys for CI and I want to tackle one problem at a time.

lentzi90 (approved):

PR #536, opened 2022-05-13
Local image build: support ubuntu

lentzi90:
Also adds documentation for how to upload, list and delete images at Artifactory.

### DIFF
--- a/ci/scripts/image_scripts/provision_base_image.sh
+++ b/ci/scripts/image_scripts/provision_base_image.sh
@@ -2,7 +2,7 @@
 
 set -eux
 
-# This script sets up fresh ubuntu 18.04+ installation
+# This script sets up fresh ubuntu 20.04+ installation
 # to create a base image for our applications.
 
 SCRIPTS_DIR="$(dirname "$(readlink -f "${0}")")"
--- a/ci/images/README.md
+++ b/ci/images/README.md
@@ -90,7 +90,7 @@ The scripts mentioned above (gen_*_image.sh) are made for use in our CI pipeline
 As such, they make certain assumptions that may not be ideal when developing and testing things locally (e.g. keypair name and ssh-key path).
 
 For these situations there is another script: `run_local.sh`.
-It allows overriding many variables and skips things like uploading the images to Artifactory.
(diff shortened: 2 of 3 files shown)


lentzi90:
/hold
I messed up when amending. There is an unbound variable. Fixing.

ci/scripts/image_scripts/provision_base_image.sh
```diff
@@ -2,16 +2,23 @@
...
+# This script sets up fresh ubuntu 20.04+ installation
 # to create a base image for our applications.
 
 SCRIPTS_DIR="$(dirname "$(readlink -f "${0}")")"
+
+# Needrestart and packer does not seem to work well together. Needrestart is
+# propmpting for what services to restart and packer cannot answer, so it get stuck.
+# This makes needrestart (l)ist the packages instead of prompting with a dialog.
+# The alternative would be sudo apt-get remove -y needrestart.
+echo '$nrconf{restart} = "l";' | sudo tee /etc/needrestart/needrestart.conf
```
lentzi90, line 14: Note this! I don't have a strong opinion on if it is better to uninstall needrestart or change the config like I did here. Went with the config change for now since it seemed less "intrusive".

lentzi90: Second note: This is only needed from ubuntu 22.04, before that the file is not there and needrestart is not installed

lentzi90:
Hmm I don't get why it cannot find the cri-o package in the test. I had no issue with it when testing with Ubuntu 22.04 and I don't see any changes in the code around it. :thinking:

lentzi90:
> Hmm I don't get why it cannot find the cri-o package in the test. I had no issue with it when testing with Ubuntu 22.04 and I don't see any changes in the code around it. thinking

Looks like the issue was that the cri-o version we were using was removed from openSUSE repositories: https://opensuse.pkgs.org/tumbleweed/opensuse-oss-x86_64/cri-o-1.23.2-1.3.x86_64.rpm.html
I bumped the version now and it seems to be working.

kashifest (approved):

PR #548, opened 2022-06-01
Add new ssh key for the CI and adding apt-get instead of apt

lentzi90:
The new ssh key uses the ed25519 format which is more secure, should
give better performance and avoid some issues with newer OSes that are
disabling some RSA hashes.

I also included a couple of small changes to make the image build go through:

- Configure needsrestart to only print kernel hints (it was hanging waiting for user input)
- Run apt-get update after apt-get dist-upgrade. This seems to be necessary or else it cannot find the packages it is trying to install.
- Use apt-get instead of apt in more places (there are warnings about using apt in scripts).

I can move these small fixes to a separate PR if required. In that case we will need to merge it before this as I was unable to build the image without them.

### DIFF
--- a/ci/scripts/image_scripts/provision_base_image.sh
+++ b/ci/scripts/image_scripts/provision_base_image.sh
@@ -9,15 +9,17 @@ SCRIPTS_DIR="$(dirname "$(readlink -f "${0}")")"
 # Needrestart and packer does not seem to work well together. Needrestart is
 # propmpting for what services to restart and packer cannot answer, so it get stuck.
 # This makes needrestart (l)ist the packages instead of prompting with a dialog.
+# It also makes it only print kernel hints instead of prompting users to acknowledge.
 # The alternative would be sudo apt-get remove -y needrestart.
-echo '$nrconf{restart} = "l";' | sudo tee /etc/needrestart/needrestart.conf || true
+echo -e '$nrconf{restart} = "l";\n$nrconf{kernelhints} = -1;' | sudo tee /etc/needrestart/needrestart.conf || true
 
 # Upgrade all packages
 sudo apt-get update
-sudo apt-get dist-upgrade -f -y                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
+sudo apt-get dist-upgrade -f -y
 
 # Install required packages.
-sudo apt install -y \
+sudo apt-get update
+sudo apt-get install -y \
   vim \
   jq \
   git \
--- a/ci/scripts/image_scripts/provision_node_image.sh
+++ b/ci/scripts/image_scripts/provision_node_image.sh
@@ -12,15 +12,17 @@ export CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-docker}"
 # Needrestart and packer does not seem to work well together. Needrestart is
 # propmpting for what services to restart and packer cannot answer, so it get stuck.
 # This makes needrestart (l)ist the packages instead of prompting with a dialog.
+# It also makes it only print kernel hints instead of prompting users to aknowledge.
 # The alternative would be sudo apt-get remove -y needrestart.
-echo '$nrconf{restart} = "l";' | sudo tee /etc/needrestart/needrestart.conf || true
+echo -e '$nrconf{restart} = "l";\n$nrconf{kernelhints} = -1;' | sudo tee /etc/needrestart/needrestart.conf || true
 
 # Upgrade all packages
 sudo apt-get update
 sudo apt-get dist-upgrade -f -y
 
 # Install required packages.
-sudo apt install -y \
+sudo apt-get update
+sudo apt-get install -y \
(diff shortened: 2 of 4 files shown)


ci/scripts/image_scripts/provision_node_image.sh
```diff
@@ -38,14 +40,14 @@ sudo apt install -y \
 sudo mv $SCRIPTS_DIR/node-image-cloud-init/retrieve.configuration.files.sh /usr/local/bin/retrieve.configuration.files.sh
 sudo chmod +x /usr/local/bin/retrieve.configuration.files.sh
 sudo apt-get install -y conntrack socat
-sudo apt install net-tools gcc linux-headers-$(uname -r) bridge-utils -y
-sudo apt install -y keepalived && sudo systemctl stop keepalived
+sudo apt-get install net-tools gcc linux-headers-$(uname -r) bridge-utils -y
```
Rozzii, line 43: How come you are using "-get" as I understand "apt" by itself is the shorter/more modern approach?

lentzi90: Apt is not yet recommended in scripts. There is a warning each time we use it actually:

> WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

The explanation for this is in the [man page](https://manpages.debian.org/buster/apt/apt.8#SCRIPT_USAGE_AND_DIFFERENCES_FROM_OTHER_APT_TOOLS). I haven't personally seen any trouble from using apt in "normal" scripts, but this openstack packer thing seems to behave in a weird way sometimes and I think using apt-get makes it better.

kashifest: I would recommend to still use apt

Rozzii: From my part after reading the man pages on the main branch of apt I see that this has not changed.
I didn't know that it is a architectural decision to make apt more of an interactive tool. From my point of view I also support the apt-get and similar action specific tools now. 
Thanks for the info @lentzi90 !

lentzi90: Ok I can put it back. But why? If there is a warning when using it?

kashifest: I had the same understanding as @Rozzii in the first comment, but if man pages say thats the case then I am ok. I thought apt is the nicer version of apt-get

ci/scripts/image_scripts/provision_base_image.sh
```diff
@@ -9,15 +9,17 @@ SCRIPTS_DIR="$(dirname "$(readlink -f "${0}")")"
 # Needrestart and packer does not seem to work well together. Needrestart is
 # propmpting for what services to restart and packer cannot answer, so it get stuck.
 # This makes needrestart (l)ist the packages instead of prompting with a dialog.
+# It also makes it only print kernel hints instead of prompting users to aknowledge.
```
kashifest, line 12: acknowledge

ci/scripts/image_scripts/provision_base_image.sh
```diff
@@ -9,15 +9,17 @@ SCRIPTS_DIR="$(dirname "$(readlink -f "${0}")")"
...
+sudo apt-get dist-upgrade -f -y
 
 # Install required packages.
-sudo apt install -y \
+sudo apt-get update
```
kashifest, line 21: this line is a duplication of line 17, we might remove either of the two?

ci/scripts/image_scripts/provision_node_image.sh
```diff
@@ -12,15 +12,17 @@ export CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-docker}"
...
-echo '$nrconf{restart} = "l";' | sudo tee /etc/needrestart/needrestart.conf || true
+echo -e '$nrconf{restart} = "l";\n$nrconf{kernelhints} = -1;' | sudo tee /etc/needrestart/needrestart.conf || true
 
 # Upgrade all packages
 sudo apt-get update
 sudo apt-get dist-upgrade -f -y
 
 # Install required packages.
-sudo apt install -y \
+sudo apt-get update
```
kashifest, line 24: same

ci/scripts/image_scripts/provision_base_image.sh
lentzi90: I thought so as well but actually the build failed without this.

The error when I didn't have this line is that it cannot find some package (jq, tree and make). Now since I knew the packages were there, I came to the conclusion that the dist-upgrade must in some way make the cache invalid. I'm not sure why, but maybe it is making some changes to the repositories/mirrors?

Anyway, I think we need both. The first one is to get fresh cache before upgrading and the second is to refresh the cache after upgrading. Does this make sense?

lentzi90: -y is only needed for install/upgrade no? The test passed like this at least :D

kashifest: ok, perhaps I am getting old , approved.

kashifest (approved):

kashifest:
Lets merge it after I solve the packer issue

PR #549, opened 2022-06-02
Uplift IPA base image to centos 9 stream

Rozzii:
This change was needed because centos 8 stream is slowly loosing upstream support and
the systemd packaged with centos 8 stream had a bug that affected udev thus the inspection
and provisioning process of Ironic.

This change also contains a additional element that puts a buildinfo file into IPA to
help with future debugging tasks.

### DIFF
--- a/ci/scripts/image_scripts/ipa_builder_elements/ipa-add-buildinfo/install.d/80-ipa-add-buildinfo
+++ b/ci/scripts/image_scripts/ipa_builder_elements/ipa-add-buildinfo/install.d/80-ipa-add-buildinfo
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+if [ "${DIB_DEBUG_TRACE:-1}" -gt 0 ]; then
+    set -x
+fi
+set -eu
+set -o pipefail
+
+echo "This IPA was built by ESJ on $(date)." > /buildinfo.txt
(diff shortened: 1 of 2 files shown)


ci/scripts/image_scripts/ipa_builder_elements/ipa-add-buildinfo/install.d/80-ipa-add-buildinfo
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+if [ "${DIB_DEBUG_TRACE:-1}" -gt 0 ]; then
+    set -x
+fi
+set -eu
+set -o pipefail
+
+echo "This IPA was built by ESJ on $(date)." > /buildinfo.txt
```
lentzi90, line 9: Will people recognize "ESJ"? Should we expand it or change it to Nordix/Metal3 or Ericsson Software Technology (Finland)?

lentzi90 (commented):
One nit below, but I will leave it up to you to decide if it should be addressed.
/lgtm

lentzi90 (approved):

ci/scripts/image_scripts/ipa_builder_elements/ipa-add-buildinfo/install.d/80-ipa-add-buildinfo
Rozzii: The use of this file is just to be able to differentiate it from other IPAs from the internet. 
You can only get this if you pull it from our Artifactory so I would not change it for this PR as it is a bit time sensitive. 
But we can have a discussion about this in private.

PR #551, opened 2022-06-02
Pin the first centos 9 based ipa

Rozzii:
This commit also contains changes that enable the artifact cleanup script
to handle the cleanup of IPA artifacts based on different base images.


Rozzii:
> Do we need to run tests on it?

I have done it already locally.

PR #552, opened 2022-06-03
Replace the old ssh key

lentzi90:
The new ssh key was already added previously as a second key in all
images. This PR removes the old key from the image building pipeline and
adds the new in its place. After merging this, only the new key will be
available on the images.

This also removes the workaround for supporting legacy crypto policies, since it is no longer needed when using the new key.


ci/images/centos_userdata.tpl
```diff
@@ -3,7 +3,6 @@ users:
   - name: ${DEFAULT_SSH_USER}
     ssh-authorized-keys:
       - ${SSH_AUTHORIZED_KEY}
-      - "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIIpxfHuI2qfTYPrL4+thyHSS78Qj9ehp2/GYxuNXthgS estjorvas@est.tech"
```
lentzi90, line 6: This may look like we are removing the new key, but what we are actually doing is putting the new key in the variable `SSH_AUTHORIZED_KEY` (see all other changes) and removing the "duplicate"

lentzi90:
Testing the metal3 images [here](https://github.com/metal3-io/project-infra/pull/398).

lentzi90:
Tests are looking good. :heavy_check_mark:

kashifest (approved):

PR #563, opened 2022-06-30
Change key name for CentOS for parallel node image building

smoshiur1237:
Parallel node image building is failing because of the same key name for Ubuntu and CentOS. This PR will change the key name for CentOS node image builiding.

### DIFF
--- a/ci/images/gen_metal3_centos_image.sh
+++ b/ci/images/gen_metal3_centos_image.sh
@@ -18,6 +18,7 @@ source "${OS_SCRIPTS_DIR}/infra_defines.sh"
 
 if [[ "$PROVISIONING_SCRIPT" == *"node"* ]]; then
   CI_IMAGE_NAME="${CI_NODE_CENTOS_IMAGE}"
+  CI_KEYPAIR_NAME="metal3ci-key-centos"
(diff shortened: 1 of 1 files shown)


ci/images/gen_metal3_centos_image.sh
```diff
@@ -61,8 +62,12 @@ echo "${REMOTE_EXEC_CMD}" > "${STARTER_SCRIPT_PATH}"
 
 # Create CI Keypair
 CI_PUBLIC_KEY_FILE="${OS_SCRIPTS_DIR}/id_ed25519_metal3ci.pub"
-delete_keypair "${SSH_KEYPAIR_NAME}"
-create_keypair "${CI_PUBLIC_KEY_FILE}" "${SSH_KEYPAIR_NAME}"
+if [[ "$PROVISIONING_SCRIPT" == *"node"* ]]; then
+  create_keypair "${CI_PUBLIC_KEY_FILE}" "${SSH_KEYPAIR_NAME}"
```
lentzi90, line 66: You need to delete it here also so it doesn't fail on the next run when the key is already there.

lentzi90 (approved):

PR #576, opened 2022-08-03
wip: Add parallel run for metal3 image building for Frankfurt and Karlskrona region

smoshiur1237:


smoshiur1237:
Nested parallel implementation is not supported by Jenkins. So this PR doesn't solve the parallel run between two region and no way to go forward. So closing it.

PR #587, opened 2022-08-28
Change node image network config to keyfile format in centos

Sunnatillo:
This PR solves issue of two connection profiles competing for eth0 interface. We set lower priority to System-eth0.nmconnection profile. 
The dev-env part of the solution is [here](https://github.com/metal3-io/metal3-dev-env/pull/1062)


kashifest (approved):

lentzi90 (approved):
LGTM! Good job!

PR #591, opened 2022-09-01
Fix package conflict

lentzi90:
- Don't install podman-catatonit. It is pulled in as a dependency to
  podman and when we explicitly install it we get conflicts with the
catatonit package.
- Use the Nordix pull through cache for container images.
- Remove some unnecessary commands:
  - podman was installed twice
  - curl and nss were updated right after updating all packages
  - one more update before installing the EPEL repo


lentzi90:
Tested manually by building the image, creating a VM from it and running `make` and `make test` in metal3-dev-env.

Rozzii (approved):

Sunnatillo (approved):
LGTM

furkatgofurov7 (approved):

PR #592, opened 2022-09-09
Use Ubuntu-22.04 as the source image and Revert MACAddressPolicy to none

mboukhalfa:
On Ubuntu 22.04 the mac address generation for Linux bridge is persistent but that makes all our nodes bridge generate the same mac address which broke our network this PR changes the mac address generation to none so the bridge will inherit the mac address of the first attached interface


kashifest:
Can you find and replace all occurences? I can see atleast one more file 
https://github.com/Nordix/metal3-dev-tools/blob/c75927204b083df097761e274fdb4594c0ed297f/ci/scripts/image_scripts/provision_base_image.sh#L5

kashifest:
/retitle Use Ubuntu-22.04 as the source image and Revert MACAddressPolicy to none

kashifest (approved):

kashifest:
Approved, would you want to run metal3-dev-tools test here or you are satisfied?

mboukhalfa:
> Approved, would you want to run metal3-dev-tools test here or you are satisfied?

ok let me run the test

PR #608, opened 2022-10-21
🐛 Exclude NetworkManager upgrades

mboukhalfa:
A workaround to the issue of losing ssh connection after creating a bridge


furkatgofurov7 (approved):
Let's give it a try!

PR #618, opened 2022-10-27
CI: Use gen_metal3_* scripts in build_node_images pipeline

mquhuy:
This is to use gen_metal3_* scripts in node images building into the same pipeline, which allows removal of gen_node_ubuntu script.


ci/images/gen_metal3_centos_image.sh
```diff
@@ -30,7 +30,7 @@ elif [[ "$PROVISIONING_SCRIPT" == *"metal3"* ]]; then
   FINAL_IMAGE_NAME="${CI_IMAGE_NAME}"
 else
   echo "Available provisioning scripts are:"
-  echo "$(ls -l ../scripts/image_scripts/provision_* | cut -f4 -d'/')"
+  echo '$(ls ../scripts/image_scripts | egrep "provision_(node|metal3)_image_centos.sh"'
```
lentzi90, line 33: This will just print the command instead of the result. Is there some problem with the old command? I can imaging issues with the relative path for example? In that case we should fix it instead I think.

```console
❯ echo '$(ls ../scripts/image_scripts | egrep "provision_(node|metal3)_image_centos.sh"'
$(ls ../scripts/image_scripts | egrep "provision_(node|metal3)_image_centos.sh"

❯ echo "$(ls -l ../scripts/image_scripts/provision_* | cut -f4 -d'/')"
provision_base_image.sh
provision_jumphost_jenkins_base_img.sh
provision_metal3_image_centos.sh
provision_metal3_image_ubuntu.sh
provision_node_image_centos.sh
provision_node_image.sh
```

ci/images/gen_metal3_ubuntu_image.sh
```diff
@@ -16,17 +17,38 @@ source "${OS_SCRIPTS_DIR}/infra_defines.sh"
...
+  BUILDER_CONFIG_FILE="image_builder_template_node.json"
+elif [[ "$PROVISIONING_SCRIPT" == *"metal3"* ]]; then
+  FINAL_IMAGE_NAME="${CI_METAL3_IMAGE}"
+  SOURCE_IMAGE_NAME="${CI_BASE_IMAGE}"
+  REMOTE_EXEC_CMD="KUBERNETES_VERSION=${KUBERNETES_VERSION} /home/${SSH_USER_NAME}/image_scripts/provision_metal3_image_ubuntu.sh"
+  IMAGE_FLAVOR="4C-16GB-50GB"
+  BUILDER_CONFIG_FILE="image_builder_template.json"
+else
+  echo "Available provisioning scripts are:"
+  echo '$(ls ../scripts/image_scripts | egrep "provision_(node|metal3)_image(_ubuntu)?.sh"'
```
lentzi90, line 46:
```suggestion
  echo "$(ls -l ../scripts/image_scripts/provision_* | cut -f4 -d'/')"
```

ci/jobs/openstack_image_building.pipeline
```diff
@@ -131,7 +136,7 @@ pipeline {
         }
       }
     }
-    stage('Build Metal3 images') {
+    stage('Build Metal3 and Node images') {
```
lentzi90, line 139: Sorry for not noticing this earlier. We cannot have the Metal3 and Node images built in the same stage, because the Metal3 image need to download the latest Node image as part of the building process. So we need to do the Node images first and then Metal3.

Build Node image -> Build Metal3 image (download the previous Node image).

We do it this way to avoid large downloads when running the tests.

lentzi90 (commented):
This will be great! Nice to get rid of the duplicated code!
Unfortunately we will still need a separate stages for Metal3 and Node images though. See comments below.

ci/images/gen_metal3_centos_image.sh
mquhuy: Ah. Thank you. I was having an impression that the purpose here is to print the command xD.
The problem with the old command is that it prints out all the provision scripts, but afaiu only the two centos scripts should be listed.

ci/images/gen_metal3_ubuntu_image.sh
mquhuy: Same goes for this.

ci/jobs/openstack_image_building.pipeline
mquhuy: Thanks a lot, I thought Node image was just a simpler version which can be used for manual tests.
So as Node image is an input of Metal3 image, it should be built in both regions then?

ci/images/gen_metal3_centos_image.sh
lentzi90: I see! That is probably not so bad, but we could change it to print `provision_*_centos.sh` so we get only the centos scripts :slightly_smiling_face:

furkatgofurov7 (dismissed):
By quick review, I noticed the node image building pipeline is deleted completely, and this is heads up, that we refer to these pipelines from [Gerrit](https://gerrit.nordix.org/admin/repos/infra/cicd,general) by their names and so without changes in Gerrit CI will fail:
to check more, `ctrl+F` inside the metal3 project: `script-path: 'ci/jobs/{id}.pipeline'  ` and check node image building job

mquhuy:
> By quick review, I noticed the node image building pipeline is deleted completely, and this is heads up, that we refer to these pipelines from [Gerrit](https://gerrit.nordix.org/admin/repos/infra/cicd,general) by their names and so without changes in Gerrit CI will fail: to check more, `ctrl+F` inside the metal3 project: `script-path: 'ci/jobs/{id}.pipeline' ` and check node image building job

Thank you for the info. Yes I believe the same pipeline in ci should be deleted as well, but I seem to lack push permission in that repo, so that's gonna be the next step to follow and both the commits should be merged at the same time.

ci/images/gen_metal3_centos_image.sh
mquhuy: Yes, that's a possibility, but it doesn't work for the ubuntu version :/ Personally I think we should use similar logic in both places.

ci/jobs/openstack_image_building.pipeline
lentzi90: No need to build them in both regions because the node images are uploaded to artifactory at the end of the pipeline. This is where the metal3 pipeline gets them from

ci/images/gen_metal3_centos_image.sh
mquhuy: Hmm. Yes, but imo it can be challenging to differentiate them if you don't know exactly what you want. The `ls -l` command even lists out the `base` and `jumphost` scripts.

With that said, if you think keeping the command as it is should be the way to go, I'm ok with it. Just wondering why :-?

ci/jobs/openstack_image_building.pipeline
mquhuy: Should we cancel the uploading/downloading to artifactory then? If we don't build node image in Fra1, the pipeline in Fra1 still needs to wait anyway, and we waste time for the upload and download steps.

ci/images/gen_metal3_centos_image.sh
lentzi90: I would keep it because if people are unfamiliar with these scripts and maybe didn't find the provisioning scripts, then it is helpful to list them so the user can easily see what they are.

ci/jobs/openstack_image_building.pipeline
lentzi90: I would do the same as for the jenkins image. Just skip building the node image in fra1.

ci/images/gen_metal3_centos_image.sh
```diff
@@ -30,7 +30,7 @@ elif [[ "$PROVISIONING_SCRIPT" == *"metal3"* ]]; then
   FINAL_IMAGE_NAME="${CI_IMAGE_NAME}"
 else
   echo "Available provisioning scripts are:"
-  echo "$(ls -l ../scripts/image_scripts/provision_* | cut -f4 -d'/')"
+  ls ../scripts/image_scripts | egrep "provision_(node|metal3)_image_centos.sh"
```
lentzi90, line 33:
```suggestion
  echo "$(ls ../scripts/image_scripts | egrep 'provision_(node|metal3)_image_centos.sh')"
```

ci/images/gen_metal3_ubuntu_image.sh
```diff
@@ -16,17 +17,38 @@ source "${OS_SCRIPTS_DIR}/infra_defines.sh"
...
+  BUILDER_CONFIG_FILE="image_builder_template_node.json"
+elif [[ "$PROVISIONING_SCRIPT" == *"metal3"* ]]; then
+  FINAL_IMAGE_NAME="${CI_METAL3_IMAGE}"
+  SOURCE_IMAGE_NAME="${CI_BASE_IMAGE}"
+  REMOTE_EXEC_CMD="KUBERNETES_VERSION=${KUBERNETES_VERSION} /home/${SSH_USER_NAME}/image_scripts/provision_metal3_image_ubuntu.sh"
+  IMAGE_FLAVOR="4C-16GB-50GB"
+  BUILDER_CONFIG_FILE="image_builder_template.json"
+else
+  echo "Available provisioning scripts are:"
+  ls ../scripts/image_scripts | egrep "provision_(node|metal3)_image(_ubuntu)?.sh"
```
lentzi90, line 46:
```suggestion
  echo "$(ls ../scripts/image_scripts | egrep 'provision_(node|metal3)_image(_ubuntu)?.sh')"
```

lentzi90 (commented):
Please echo the output also

ci/jobs/openstack_image_building.pipeline
mquhuy: I'm actually not sure how it would go in that case. Jenkins image build happens at the end of the pipeline, so we can skip it however we want, but if we skip node build here, we would need to ask metal3 build in fra1 to wait for the node build in kna1 to finish first, otherwise it would just build immediately and get the old node image as base.

Of course we can just put node building to a separate stage, I'm just wondering if that makes senses to keep the upload-download steps. As fra1 has to wait, we can as well build node image there in parallel with kna1, then it can use the one it's just built. Another option is to keep the node image pipeline, and only update it to use the metal3 script.

lentzi90: This is how I imagine it

```


                                                                                      ┌─────►Metal3 image ubuntu Fra1│
                                                                                      │                              │
                                                                                      │                              │
                                                                                      │                              │
       ┌──────►Base image Kna1────────►Jenkins image Kna1───────►Node image ubuntu│   ├─────►Metal3 image ubuntu Kna1│
       │                                                                          │   │                              │
 Start │                                                                          ├──►│                              ├──────►End
       │                                                                          │   │                              │
       └──────►Base image Fra1────────►Skip              ───────►Node image centos│   ├─────►Metal3 image centos Fra1│
                                                                                      │                              │
                                                                                      │                              │
                                                                                      │                              │
                                                                                      └─────►Metal3 image ubuntu Fra1│


```
So build the node images as a separate stage after the jenkins image.
In theory we could do the node images in parallel with the jenkins build but I'm not sure if it is worth the extra complexity.

ci/images/gen_metal3_centos_image.sh
mquhuy: Yeah, that's what I thought as well, but I think that's the point of this egrep: it lists only relevant options, so that those who're unfamiliar are less confused.

ci/jobs/openstack_image_building.pipeline
mquhuy: Yeah, I don't have a strong opinion, but in this case the node image build time is added to the openstack build time without any apparent benefits, so I'm wondering if it's worth it. Currently node image builds only happen once every 15 days, but with this in place it will happen every day.

Anw you comment made me think of another scenario: what if we build the node images in Fra1 while Kna1 is building the jenkins image? Will that save us time? Otherwise I would vote to keep the current node pipeline (which runs every 15 days), and change it to use the gen_metal3 script and matrix.

lentzi90: > Otherwise I would vote to keep the current node pipeline (which runs every 15 days), and change it to use the gen_metal3 script and matrix.

This sounds good. +1 from me :slightly_smiling_face:

ci/images/gen_metal3_ubuntu_image.sh
```diff
@@ -56,8 +79,14 @@ packer build \
...
+  "${IMAGES_DIR}/${BUILDER_CONFIG_FILE}"
 
 # Replace any old image
 
 replace_image "${IMAGE_NAME}" "${FINAL_IMAGE_NAME}"
+
+# upload node image to artifactory
+if [[ "$PROVISIONING_SCRIPT" == *"node"* ]]; then
+  bash "${SCRIPTS_DIR}/upload_node_image_rt.sh" "${FINAL_IMAGE_NAME}"
+fi
```
Rozzii, line 92: Empty line at EOF is missing , please add one.

Rozzii (dismissed):
Looks good to me just one minor nit.

ci/images/gen_metal3_ubuntu_image.sh
mquhuy: Thanks. Added now.

ci/jobs/openstack_node_image_building.pipeline
```diff
@@ -49,56 +49,39 @@ pipeline {
...
-                      provision_node_image_centos.sh"
-                  }
+        options {
+          timeout(time: 60, unit: 'MINUTES')
+        }
+        stages {
+          stage('Building node image'){
+            environment {
+              TEST_EXECUTER_VM_NAME = "${TEST_EXECUTER_VM_NAME}-${IMAGE_OS}-based"
+              PROVISION_SCRIPT = "provision_node_image${env.IMAGE_OS == "CentOS" ? "_centos" : ""}.sh"
```
lentzi90, line 67: What if we just rename this script to make it simpler? `provision_node_image_ubuntu.sh` and `provision_node_image_centos.sh`. Easy to understand and simpler to use IMO :slightly_smiling_face:

mquhuy: Thanks, that's absolutely much better, but I wasn't sure if this script was referenced somewhere else that I'm not aware of so I opted for the safe option xD Have you noticed it referenced somewhere else?

lentzi90: I'm quite sure this is the only place, so let's do it! If it breaks something we fix it :smile:

mquhuy: Yep. Let's do it then :D

PR #620, opened 2022-11-09
Configure nameservers for the metal3 ubuntu image

lentzi90:
Testing cloudflare DNS here. On centos we use google. Any of them should be better than the uncertainty and unreliability due to the DHCP added nameservers in openstack.

Also updated docs to explain how to create an openstack keypair.


lentzi90:
Sorry, updated nameservers to cloudflare.
/test-metal3-fullstack

tuminoid:
LGTM

PR #629, opened 2022-11-28
Use apt-get instead of apt in scripts

tuminoid:
Apt does not have stable CLI, and gives out warnings when used from script context. Replace apt with apt-get.


lentzi90 (approved):
Great cleanup, thanks!

PR #646, opened 2023-01-10
add support for custom config-drive in IPA part 2

Rozzii:
This commit introduces the following changes:
    
        - New ipa-file-injector service that can inject files
          from config-drive with a custom label during userspace boot.
        - ipa-file-injector expects a manifest to be present on the
          config-drive to contain key-value pairs in a ":" separated
          format that specify source and destination paths for the
          files that are planned to be copied.
        - Custom label used by the ipa-file-injector can be specified during
          the IPA build process.


Rozzii:
/hold
Part 1 has to be merged first, then I will rebase this.

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/install.d/99-enable-ipa-file-injector
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+if [ ${DIB_DEBUG_TRACE:-1} -gt 0 ]; then
+    set -x
+fi
+set -eu
+set -o pipefail
```
tuminoid, line 7: No pipes, not needed.

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/install.d/99-enable-ipa-file-injector
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+if [ ${DIB_DEBUG_TRACE:-1} -gt 0 ]; then
+    set -x
+fi
+set -eu
```
tuminoid, line 6: The only variable -u could take effect is above it. Maybe move it to the top. set -e also makes most sense when its at very start.

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/99-install-file-injector
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+if [ ${DIB_DEBUG_TRACE:-1} -gt 0 ]; then
+    set -x
+fi
+set -eu
+set -o pipefail
+
+[ -n "$TARGET_ROOT" ]
```
tuminoid, line 9: This would be 100% better if you exit with a message, than just exit because it fails and -e kicks you out. Also, missing braces.

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+set -eu
+set -o pipefail
+
+# Behaviour of the scritp is the following:
+# Try and mount the config-drive; if it exists, then copy files
+# to locations as specified in the file_injection.manifest of the
+# config-drive.
+
+function copy_files_from_drive {
```
tuminoid, line 10: With explicit bash, `function x {` is deprecated, and should use `x() {` style functions.

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
```diff
@@ -0,0 +1,50 @@
...
+    local MOUNT_POINT="${1}"
```
tuminoid, line 11: While braces work with `${1}` etc, it is best-practice for single character variables to use plain `$1`, unless you're doing some expansion for them.

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
```diff
@@ -0,0 +1,50 @@
...
+    fi
+fi
+
+# Mount config drive
+mkdir -p /mnt/config
+BLOCKDEV="$(blkid -L ${CONFIG_DRIVE_LABEL})"
+TYPE="$(blkid -t LABEL=${CONFIG_DRIVE_LABEL} -s TYPE -o value)"
+
+# Mounting won't fail if the mount already exists
+if [[ "${TYPE}" == 'vfat' ]]; then
```
tuminoid, line 40: You're using `[` in previous if, and now `[[` which is not necessary. Unless you use bash-specific advanced comparison, like regex, single = and single [ is the way to go.

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
```diff
@@ -0,0 +1,50 @@
...
+    mount -t vfat -o umask=0077 "${BLOCKDEV}" "${MOUNT_POINT}" || true
```
tuminoid, line 41: What's the point of `|| true`, since it'll just defer failure to `copy_files_from_drive` and make it harder to debug?

ci/scripts/image_scripts/ipa_builder_elements/override-simple-init/post-install.d/99-override-simple-init
```diff
@@ -0,0 +1,12 @@
+if [ ${DIB_DEBUG_TRACE:-1} -gt 0 ]; then
+    set -x
+fi
+set -eu
+set -o pipefail
```
tuminoid, line 5: Same nit as earlier about the set lines.

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
```diff
@@ -0,0 +1,50 @@
...
+# Try and mount the config-drive; if it exists, then copy files
+# to locations as specified in the file_injection.manifest of the
+# config-drive.
+
+function copy_files_from_drive {
+    local MOUNT_POINT="${1}"
+    local MANIFEST="${2}"
+    while IFS=":" read -r src dst; do
+        cp -r "${MOUNT_POINT}/${src}" "${dst}"
+    done <"${MANIFEST}"
```
tuminoid, line 15: Add a space after `<` for readability.

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
tuminoid: This function is also used exactly once, so it doesn't really help. Probably would be earlier to read, and less lines, if it was inline where function is called.

tuminoid (changes requested):

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/install.d/99-enable-ipa-file-injector
Rozzii: Thanks, fixed!

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/install.d/99-enable-ipa-file-injector
Rozzii: Thanks, fixed!

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
Rozzii: Thanks, fixed!

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/99-install-file-injector
Rozzii: Thanks, fixed!

ci/scripts/image_scripts/ipa_builder_elements/override-simple-init/post-install.d/99-override-simple-init
Rozzii: Thanks, fixed!

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
Rozzii: Thanks, fixed!

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
Rozzii: Thanks, fixed!

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
Rozzii: I don't think it is hard to debug, there will be a error message saved related to the read command on the ipa-file-injector service' status page. It can be checked with systemctl status.

Rozzii: But I will implement something a bit more sophisticated some pre-check for example why not.

Rozzii: Done.
I have modified the error handling process.

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/install.d/99-enable-ipa-file-injector
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+if [ "${DIB_DEBUG_TRACE:-1}" -gt 0 ]; then
+    set -x
+fi
+set -e
```
tuminoid, line 6: Add `-u` back, and move to the top of the script.

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/99-install-file-injector
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+if [ "${DIB_DEBUG_TRACE:-1}" -gt 0 ]; then
+    set -x
+fi
+set -e
```
tuminoid, line 6: Same here

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/99-install-file-injector
```diff
@@ -0,0 +1,25 @@
...
+
+if [ -z "$TARGET_ROOT" ]; then
```
tuminoid, line 8: Missing braces.

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
```diff
@@ -0,0 +1,67 @@
...
+    local MOUNT_POINT="$1"
+    local MANIFEST="$2"
+    local RETURN_VAL="0"
+    local CPY_RESULT="0"
+    if ! [ -f "${MANIFEST}" ]; then
+	echo "ERROR: The config-drive's file injection manifest is not present here: ${MANIFEST}"
+        exit 1
+    fi
+    while IFS=":" read -r src dst; do
+        cp -r "${MOUNT_POINT}/${src}" "${dst}"
```
tuminoid, line 18: You can write `if ! cp -r ...` directly, saves you trouble of assigning CPY_RESULT, and if clause.  It also prevents script from failing when `set -e` is active (which it isn't, you should add `set -eu` on top, since if catches the failure.

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+
```
tuminoid, line 2: Add `set -eu` on top.

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
```diff
@@ -0,0 +1,67 @@
...
+if [ -z "${BLOCKDEV}" ]; then
+    echo "ERROR: The block device with the ${CONFIG_DRIVE_LABEL} label can't be found!"
+    exit 1
+fi
+TYPE="$(blkid -t LABEL=${CONFIG_DRIVE_LABEL} -s TYPE -o value)"
+
+# Mounting won't fail if the mount already exists
+# If the mount didn't exists at all it is expected to fail in the copy_files_from_drive function
+if [ "${TYPE}" == 'vfat' ]; then
+    mount -t vfat -o umask=0077 "${BLOCKDEV}" "${MOUNT_POINT}"
```
tuminoid, line 58: I see you removed `|| true` here, but at the cost of removing `set -e`. `set -e` is always better choice than cleaning up `|| true`. As you explained in the comment, you prefer to defer failure to the copy_files function, so its design choice, and its commented as well.

ci/scripts/image_scripts/ipa_builder_elements/override-simple-init/post-install.d/99-override-simple-init
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+if [ "${DIB_DEBUG_TRACE:-1}" -gt 0 ]; then
+    set -x
+fi
+set -e
```
tuminoid, line 6: Same here.

tuminoid (changes requested):
Some nits, but also one thing to fix.

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
Rozzii: I am conciously not using set -e in this script, i want this to fail only at those locations where I have implemented error handling, if it fails in other places like blkid commands or mount, then that is fine from the perspecitve of this service those are tolerated, it is also fine if some or all of the copy fails, missing files are also tolerated the script will fail at the end but I want it to try to execute all to copy actions.

I can do the if cp -r  I agree that is a bit more lean.

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
Rozzii: It is  by design, only specific command failures are considered service failure.

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
Rozzii: It is by design, only specific command failures are considered service failure.

but I see now you prefer the explicit || true so I will turn the things around add add back the || true  .

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
Rozzii: I see your point now, I will change it.

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
Rozzii: I will change the set -eu back too, I see your point .

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/install.d/99-enable-ipa-file-injector
Rozzii: done

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/99-install-file-injector
Rozzii: done

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/99-install-file-injector
Rozzii: done

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
Rozzii: done

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
Rozzii: done

ci/scripts/image_scripts/ipa_builder_elements/override-simple-init/post-install.d/99-override-simple-init
Rozzii: done

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
Rozzii: done

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/99-install-file-injector
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+set -eu
+if [ "${DIB_DEBUG_TRACE:-1}" -gt 0 ]; then
+    set -x
+fi
+
+if [ -z "${TARGET_ROOT}" ]; then
```
tuminoid, line 8: Now that we have `set -u` this will fail if TARGET_ROOT is undefined. If only triggers the error if TARGET_ROOT is empty. You want to have `if [ -z "${TARGET_ROOT:-}" ]; ...`. Sorry for missing this on previous round.

tuminoid (changes requested):

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/99-install-file-injector
Rozzii: I think it is fine if it is a hard failure here, TARGET_ROOT is a must have in DIB, if this is missing I think it is ok to fail as that would mean that there is some issue with the base os element. Also the TARGET_ROOT is dynamically generated so there is no default value for it.

tuminoid: OK, in that case it might warrant a comment to say it is intentional. Helps the next guy who has to debug it to know where the fault lies.

Rozzii: Done.

ci/scripts/image_scripts/ipa_builder_elements/ipa-file-injector/root.d/ipa-file-injector.sh
```diff
@@ -0,0 +1,68 @@
...
+# Behaviour of the scritp is the following:
+# Try and mount the config-drive; if it exists, then copy files
+# to locations as specified in the file_injection.manifest of the
+# config-drive.
+
+copy_files_from_drive() {
+    local MOUNT_POINT="$1"
+    local MANIFEST="$2"
+    local RETURN_VAL="0"
+    if ! [ -f "${MANIFEST}" ]; then
```
tuminoid, line 13: Typically the `!` is inside the `[]`, but it works like this as well. Semantics about is it the `[` (ie. `test` command) flipping the result of a test, or `if` flipping the return value of `[`.

tuminoid (approved):
One comment on semantics, but it works both ways, so lgtm on scripting pov.

PR #648, opened 2023-01-17
add retries to apt and packer

Rozzii:
This commit adds retires to apt and packer commands involved in openstack,jumphost,base and node image building.

### DIFF
--- a/ci/images/image_builder_template.json
+++ b/ci/images/image_builder_template.json
@@ -35,11 +35,13 @@
     {
       "type": "file",
       "source": "{{user `local_scripts_dir`}}",
-      "destination": "/home/{{user `ssh_username`}}/"
+      "destination": "/home/{{user `ssh_username`}}/",
+      "max_retries": 5
     },
     {
       "type": "shell",
-      "script": "{{user `exec_script_path`}}"
+      "script": "{{user `exec_script_path`}}",
+      "max_retries": 5
     }
   ]
 }
--- a/ci/images/image_builder_template_node.json
+++ b/ci/images/image_builder_template_node.json
@@ -38,11 +38,13 @@
       {
         "type": "file",
         "source": "{{user `local_scripts_dir`}}",
-        "destination": "/home/{{user `ssh_username`}}/"
+        "destination": "/home/{{user `ssh_username`}}/",
+        "max_retries": 5
       },
       {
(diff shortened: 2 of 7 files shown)


ci/scripts/image_scripts/setup_chrony_ubuntu.sh
```diff
@@ -1,5 +1,9 @@
 #! /usr/bin/env bash
 
+# Set apt retry limit to higher than default to
+# make the data retrival more reliable
+sudo echo 'Acquire::Retries "10";' > /etc/apt/apt.conf.d/80-retries
```
tuminoid, line 5: Don't put any code above `set -eu` as it might fail and set wouldn't catch it.

tuminoid (changes requested):

ci/scripts/image_scripts/setup_chrony_ubuntu.sh
Rozzii: there is no set -eu in the script

Rozzii: done

Rozzii:
> Do we really need to put the apt config in all these scripts? Is it not enough to put it in `provision_*.sh` scripts that then call the other scripts?

I was not sure about all teh use-cases of all the script, just wanted to be sure that the retries are there, if you think it is not needed everywhere i can put it only to provision scripts

tuminoid (approved):
lgtm

lentzi90:
> > Do we really need to put the apt config in all these scripts? Is it not enough to put it in `provision_*.sh` scripts that then call the other scripts?
> 
> I was not sure about all the use-cases of all the scripts, just wanted to be sure that the retries are there, if you think it is not needed everywhere i can put it only to provision scripts

I'm fairly certain that they are only ever used by the provision scripts. Did a quick search through the repo for any references and could not find any other than those scripts.
Strictly speaking the metal3 and jenkins images are also built on the base image so we would only need it in base and node provision scripts, but that could change in the future. I would anyway suggest to add it only in the provision scripts.

kashifest:
> /retest

I am fixing the shellcheck issues in my new PR. It wont pass until that merges

kashifest (approved):

PR #654, opened 2023-02-10
fix run_local image builder script

Rozzii:
- fix failing parameter expension of the container env vars (this was caused by shellcheck related refactoring)

- remove unnecessary container env vars

- add timestamp to the generated image otherwise in case of

  subsequent image builds the openstack client won't be abel to distinguish between the images and won't save the new image

- introduce more explicit variable nameing in certain cases

- remove key_pair variable from the packer vars as it is not needed and just complicates the script

- restrict openstack related logic into the container, so nothing that talks to openstack will be executed outside the container thus there is no need to have the ostack client installed on the host system

- add sandbox mode, in case the user would like to look around in ostack without building an image

- add interactive debug mode for the packer building process, user can wait between packer build steps thus the build vm won't disappear in case of an issue

Co-authored-by: Lennart Jern lennart.jern@est.tech

### DIFF
--- a/ci/images/README.md
+++ b/ci/images/README.md
@@ -105,14 +105,16 @@ Here is an example of custom variables:
 
 ```bash
 # My custom variables
-export SSH_USER_NAME="foo"
-export SSH_USER_GROUP="wheel"
-export SSH_KEYPAIR_NAME="foo-key"
-export SSH_PUBLIC_KEY_FILE="/home/foo/.ssh/id_rsa.pub"
-export SSH_PRIVATE_KEY_FILE="/home/foo/.ssh/id_rsa"
+export PACKER_SSH_USER_NAME="foo"
+export USERDATA_SSH_PUBLIC_KEY_FILE="/home/foo/.ssh/id_rsa.pub"
+export PACKER_SSH_PRIVATE_KEY_FILE="/home/foo/.ssh/id_rsa"
 export IMAGE_NAME="foo-test-image"
 export SOURCE_IMAGE_NAME="Upstream-centos-or-ubuntu"
-export NETWORK="abcdefg"
+# Openstack network name, vm will receive Openstack internal IP from this network.
+export OS_NETWORK_NAME="ext-net"
+# If user wants to use Openstack network ID instead of the network name,
+# then this variable can be set this instead (then network ID lookup won't happen):
+export OS_NETWORK_ID="abcdefg"
 # Artifactory variables
 # These are needed if you want to build Node images and upload them to Artifactory
 export RT_URL="https://artifactory.nordix.org/artifactory"
@@ -123,6 +125,19 @@ export RT_TOKEN=<your-token>
 **NOTE:** The script uploads the node images (i.e. those produced by `provision_node_image_ubuntu.sh` and `provision_node_image_centos.sh`) to Artifactory.
 Make sure you use an image name that does not conflict with existing images!
 This is also a good idea for other images since they will end up in Openstack.
+
+Additional configuration options:
+```bash
+export PACKER_DEBUG_ENABLED="true/false" # if true runs the packer build in interactive debug mode,
+                                         # packer stops between build stages
+export IMAGE_CLEANUP="true/false" # if true deletes the image from openstack that has an equivalent
+                                  # name with the one specifed in the IMAGE_NAME variable, if false doesn't delete any image,
+                                  # instead appends a timestamp to the name of each newly built image to avoid collision
+```
+
+If the the `./run_local sandbox` command is issued the packer build and the artifactory upload process
+won't be executed, rather the user will be presented with an interactive container environment where
+the openstack cli can be used to check the openstack resources.
 
 And here is how to use it:
 
--- a/ci/images/run_local.sh
+++ b/ci/images/run_local.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # This script can be used to run locally one of the image building
 # scripts. It reproduces the CI environment.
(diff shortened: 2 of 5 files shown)


ci/images/run_local.sh
```diff
@@ -67,69 +76,80 @@ else
...
-# If this is not the case, set the NETWORK directly instead.
-NETWORK="${NETWORK:-$(get_resource_id_from_name network "${NETWORK_NAME}")}"
+# If NETWORK is not set directly, try to get it based on NETWORK_NAME.
+if [ -z "${NETWORK}" ]; then
+  NETWORK=$("${CONTAINER_RUNTIME}" run --rm \
+            "${CR_CMD_ENV[@]}" \
+	    "-v" "${DEV_TOOLS}":"/data/metal3-dev-tools" \
+	    "registry.nordix.org/metal3/image-builder" \
+            bash -c "source ${CONTAINER_OS_UTILS} && get_resource_id_from_name network ${NETWORK_NAME}")
+fi
```
lentzi90, line 117: Nice!

ci/images/run_local.sh
```diff
@@ -19,30 +19,39 @@ set -eux
...
+USERDATA_SSH_USER_GROUP="${USERDATA_SSH_USER_GROUP:-wheel}"
 # The network name is by defualt set to $DEV_EXT_NET. You can override it here.
-NETWORK_NAME=""
+NETWORK_NAME="${NETWORK_NAME:-}"
 # Or if you want to use ID, set this instead:
-# NETWORK=<ID-of-network>
+NETWORK="${NETWORK:-}"
+# IMAGE_DATE is used to distinguish between images because the openstack client
+# can't download an image if there are multiple images present in OS with the same name
+IMAGE_DATE="$(date +"%Y-%m-%dT%H-%M-%S%z")"
```
lentzi90, line 46: While it is nice to avoid name collisions, I do like to be able to set the exact name when I build an image, but I can always rename them of course. Another thing to think about is that this makes it easier to forget about old images when you don't have to remove them before building a new with the same name. Also, deleting them becomes a bit harder when you have to check the exact name to get the date correct.
What do you think about checking if the image name is already taken before starting the build? I don't have a strong opinion on this myself. Adding the date is fine, just wanted to raise some drawbacks and discuss alternatives. :slightly_smiling_face:

lentzi90 (changes requested):
Nice job! This is definitely improving many things!

Could you also update the readme also so it matches?
https://github.com/Nordix/metal3-dev-tools/blob/main/ci/images/README.md?plain=1#L87-L175

ci/images/run_local.sh
```diff
@@ -19,30 +19,39 @@ set -eux
...
 #
 # Overridable variables
 #
 # These variables can be overridden by the user, but have defaults that may or
 # may not be useful.
 CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-docker}"
-SSH_USER_NAME="${SSH_USER_NAME:-${USER}}"
-SSH_PUBLIC_KEY_FILE="${SSH_PUBLIC_KEY_FILE:-/home/${USER}/.ssh/id_rsa.pub}"
-SSH_PRIVATE_KEY_FILE="${SSH_PRIVATE_KEY_FILE:-/home/${USER}/.ssh/id_rsa}"
+KUBERNETES_VERSION=${KUBERNETES_VERSION:-"v1.26.0"}
```
tuminoid, line 28: Quotes go around the whole expression, not needed inside the braces.

ci/images/run_local.sh
```diff
@@ -67,69 +76,80 @@ else
...
 DEV_TOOLS="$(dirname "$(readlink -f "${0}")")/../../"
-SCRIPTS_DIR="${DEV_TOOLS}/ci/scripts/image_scripts"
 OS_SCRIPTS_DIR="${DEV_TOOLS}/ci/scripts/openstack"
 IMAGES_DIR="${DEV_TOOLS}/ci/images"
 
-# shellcheck source=ci/scripts/openstack/infra_defines.sh
-source "${OS_SCRIPTS_DIR}/infra_defines.sh"
-# shellcheck source=ci/scripts/openstack/utils.sh
-source "${OS_SCRIPTS_DIR}/utils.sh"
+CR_CMD_ENV=("--env" "OS_AUTH_URL"
```
tuminoid, line 82: Please put each "--env ..." on separate line, including the first and last one. Makes diffs cleaner if touched later on.

ci/images/run_local.sh
```diff
@@ -67,69 +76,80 @@ else
...
+  "--env" "OS_PROJECT_NAME"
+  "--env" "OS_TENANT_NAME"
+  "--env" "OS_AUTH_VERSION"
+  "--env" "OS_IDENTITY_API_VERSION"
+  "--env" "OS_USERNAME"
+  "--env" "OS_PASSWORD")
+
+
+if [ "${PROVISIONING_SCRIPT}" = "sandbox" ]; then
+  "${CONTAINER_RUNTIME}" "run --rm -ti"\
```
tuminoid, line 95: `"run --rm -ti"` is interpreted as single word, and won't work. No need to quote them.

ci/images/run_local.sh
```diff
@@ -139,8 +159,17 @@ CONTAINER_IMAGES_DIR="/data/metal3-dev-tools/ci/images"
     -var "flavor=${IMAGE_FLAVOR}" \
     "${CONTAINER_IMAGES_DIR}/${BUILDER_CONFIG_FILE}"
 
+if [[ "$PROVISIONING_SCRIPT" == *"node"* ]]; then
```
tuminoid, line 162: Braces.

ci/images/run_local.sh
```diff
@@ -139,8 +159,17 @@ CONTAINER_IMAGES_DIR="/data/metal3-dev-tools/ci/images"
     -var "flavor=${IMAGE_FLAVOR}" \
     "${CONTAINER_IMAGES_DIR}/${BUILDER_CONFIG_FILE}"
 
+if [[ "$PROVISIONING_SCRIPT" == *"node"* ]]; then 
+  
+  CR_CMD_ENV+=("--env" "RT_URL"
```
tuminoid, line 164: Each value on separate line.

tuminoid (changes requested):
Some nits. One error found. Change the shebang to `#!/usr/bin/env bash` as well. `run_local.sh` is specifically meant to be executed on user's machine.

ci/images/run_local.sh
Rozzii: done.

ci/images/run_local.sh
Rozzii: done.

ci/images/run_local.sh
Rozzii: done

ci/images/README.md
```diff
@@ -105,14 +105,14 @@ Here is an example of custom variables:
...
-export SSH_PUBLIC_KEY_FILE="/home/foo/.ssh/id_rsa.pub"
-export SSH_PRIVATE_KEY_FILE="/home/foo/.ssh/id_rsa"
+export PACKER_SSH_USER_NAME="foo"
+export USERDATA_SSH_PUBLIC_KEY_FILE="/home/foo/.ssh/id_rsa.pub"
+export PACKER_SSH_PRIVATE_KEY_FILE="/home/foo/.ssh/id_rsa"
 export IMAGE_NAME="foo-test-image"
 export SOURCE_IMAGE_NAME="Upstream-centos-or-ubuntu"
 export NETWORK="abcdefg"
+or
+export NETWORK_NAME="ext-net"
```
tuminoid, line 115: It would be helpful if difference between these two and expected content would be clearer here for both `NETWORK` and `NETWORK_NAME`.

ci/images/run_local.sh
Rozzii: done

ci/images/README.md
```diff
@@ -124,6 +124,19 @@ export RT_TOKEN=<your-token>
 Make sure you use an image name that does not conflict with existing images!
 This is also a good idea for other images since they will end up in Openstack.
 
+Additional configuration options:
+```bash
+export PACKER_DEBUG_ENABLED="true/false" # if true runs the packer build in interactive debug mode,
+packer stops between build stages
+export NO_IMAGE_CLEANUP="true/false" # if ture deletes the image from openstack that has an equivalent
```
tuminoid, line 131: ture -> true

ci/images/README.md
```diff
@@ -124,6 +124,19 @@ export RT_TOKEN=<your-token>
...
+name with the one specifed in the IMAGE_NAME variable, if false doesn't delete any image, instead
+appends a timestamp to the name of each newly built image to avoid collision
```
lentzi90, line 133: Continuing the comments on the next line looks strange since it is not syntax highlighted in the same way. It also makes it harder to copy this snippet. I would suggest putting the comments on their own line(s) completely or at least add a `#` on the continuing lines.

ci/images/run_local.sh
lentzi90: Nice!
Can we drop the `NO_` from `NO_IMAGE_CLEANUP`? It is easier to understand and reason about it if the variable in itself doesn't have extra negations IMO. :sweat_smile: So just `IMAGE_CLEANUP="true/false"

Rozzii: Sure, done.

ci/images/README.md
Rozzii: added some explanation and made the names a bit more explicit.

ci/images/README.md
Rozzii: done.

ci/images/README.md
Rozzii: done.

ci/images/README.md
```diff
@@ -124,6 +126,19 @@ export RT_TOKEN=<your-token>
 Make sure you use an image name that does not conflict with existing images!
 This is also a good idea for other images since they will end up in Openstack.
 
+Additional configuration options:
+```bash
+export PACKER_DEBUG_ENABLED="true/false" # if true runs the packer build in interactive debug mode,
+                                         # packer stops between build stages
+export IMAGE_CLEANUP="true/false" # if false deletes the image from openstack that has an equivalent
+                                  # name with the one specifed in the IMAGE_NAME variable, if true doesn't delete any image, instead
+                                  # appends a timestamp to the name of each newly built image to avoid collision
```
lentzi90, line 135: One last thing, I think we have true/false backwards here. IMAGE_CLEANUP=true should delete, false should add the timestamp

tuminoid (approved):
Looks good from my pov

tuminoid:
/hold
Let's make sure we have all the reviews in.

ci/images/README.md
Rozzii: Just mixed up the description, nice find.

lentzi90 (approved):
/lgtm

PR #674, opened 2023-05-08
re-enable cloud-init on centos node images

Rozzii:
This commit is a fix to the following issue:
In the past cloud-init based network configuration was disabled on centos node images during build for some reason. Until cloud-init network config process is disabled IPAM based network configuration is also blocked thus IPAM integration is not tested during any integration tests that are using centos node images.


Sunnatillo (commented):
/lgtm

As far as remember it was done because, cloud init would mess up resolv conf configuration. It seems the issue is gone now.

tuminoid:
/retitle re-enable cloud-init on centos node images

lentzi90:
> I am not an expert so I have to ask, how does using the same name server as the one set in the cleura subnet could cause an issue? I mean in practice it means that both the cleura subnet and the node image is using Google dns, why would that be an issue?

It was the nameserver itself which was the issue. We got lookup issues which broke the tests simply :slightly_smiling_face: 

> The dns server used in dev-env will be 8.8.8.8 after the removal of the code in this PR, this comes from the metal3DataTemplate of dev-env.

Great!
/hold cancel

tuminoid (approved):

lentzi90:
> The dns server used in dev-env will be 8.8.8.8 after the removal of the code in this PR, this comes from the metal3DataTemplate of dev-env.

Just realized this will of course only apply to nodes, not the VM that runs the test. Anyway I think we can handle any issues through Cleura.

lentzi90 (approved):

PR #682, opened 2023-06-26
Verify node image is not corrupted

Sunnatillo:
This PR adds a check where newly built node image is checked for corruptions by converting it to IMG disk image format before pushing it to artifactory

### DIFF
--- a/ci/scripts/image_scripts/upload_node_image_rt.sh
+++ b/ci/scripts/image_scripts/upload_node_image_rt.sh
@@ -12,7 +12,7 @@
 #   ./upload_node_image_rt.sh CENTOS_8.2_NODE_IMAGE_K8S_v1.18.8
 # Don't include the image format, only the name.
 # Example: CENTOS_8.2_NODE_IMAGE_K8S_v1.18.8 instead of CENTOS_8.2_NODE_IMAGE_K8S_v1.18.8.qcow2
-set -x
+set -eux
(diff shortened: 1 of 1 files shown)


Sunnatillo:
Worker still had old container image

lentzi90 (dismissed):
/approve

ci/scripts/image_scripts/upload_node_image_rt.sh
```diff
@@ -24,13 +24,19 @@ WORK_DIR=/tmp/node_image
 mkdir -p "$WORK_DIR"
 echo "Downloading node Image from openstack"
 openstack image save --file "${WORK_DIR}"/"$IMAGE_NAME".qcow2 "$IMAGE_NAME"
-IMAGE_NAME="$IMAGE_NAME".qcow2
+
+# Verify the image for any signs of corruption before pushing it to artifactory
+echo "Verify the image for any signs of corruption by converting it to the IMG format"
```
tuminoid, line 29: This does not actually matter since the script is not run with `set -e`, so it might fail but it just keeps going.

Change the `set -x` to `set -eux` to get the effect you're looking for.

Sunnatillo: Thank you for review. 
Fixed now

lentzi90 (approved):

tuminoid (approved):
/lgtm

PR #683, opened 2023-06-27
Add qemu-img package to image-builder container image

Sunnatillo:
Needed for #682


lentzi90 (approved):
/approve
Noting here also that `wget` was duplicated (so not removed here even though it looks like it)

tuminoid (approved):
/lgtm

PR #693, opened 2023-08-29
Fix image sharing stage

Sunnatillo:


Sunnatillo:
> /approve
> 
> Just to double check: The `IMAGE_OS` var was removed because it is set from the `axis`, right?

yes. IT was leftover from test I was doing. For clarity it is removed now.

kashifest (approved):
/lgtm

ci/scripts/image_scripts/share_image.sh
```diff
@@ -1,4 +1,5 @@
-#!/usr/bin/env bash
+#! /usr/bin/env bash
```
tuminoid, line 1: FYI: This space is not necessary, nor is the empty line below it. Space usually is not added, while the empty line often is. Doesn't matter either way.

ci/jobs/openstack_image_building.pipeline
```diff
@@ -183,7 +182,14 @@ pipeline {
                 withCredentials([
                   usernamePassword(credentialsId: 'metal3ci_city_cloud_openstack_credentials', usernameVariable: 'OS_USERNAME', passwordVariable: 'OS_PASSWORD')
                 ]) {
-                    sh "/data/ci/scripts/image_scripts/share_image.sh"
+                  catchError([stageResult: 'FAILURE', message: "Failed to share the Metal3 ${IMAGE_OS} image for dev2 project ${REGION} region"]) {
+                    sh "docker run --rm \
+                    ${DOCKER_CMD_ENV}\
+                    --env IMAGE_OS \
+                    -v ${CURRENT_DIR}:/data \
```
tuminoid, line 189: This will break of CURRENT_DIR has a space.

We probably have plenty of the same error elsewhere, so fixing them could be done separately.

PR #697, opened 2023-09-04
Prevent NetworkManager to change /etc/resolv.conf

Sunnatillo:
This PR disables NetworKManager resolv.conf file changes. Instead adds dns addresses manually for centos images.

### DIFF
--- a/ci/scripts/image_scripts/configure_network_centos.sh
+++ b/ci/scripts/image_scripts/configure_network_centos.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+# The goal of this script is to prevent NetworkManager set
+# dns server addresses on /etc/resolve.conf 
+
+# make NetworkManager to prevent modifying resolv.conf
+sudo sed -i '/^\[main\]/a dns=none' /etc/NetworkManager/NetworkManager.conf
+sudo systemctl restart NetworkManager.service
+
+# Specify DNS servers for systemd-resolved.
+cat <<EOF | sudo tee /etc/resolv.conf
+nameserver 1.1.1.1
+nameserver 8.8.8.8
+
+EOF
--- a/ci/scripts/image_scripts/configure_network_ubuntu.sh
+++ b/ci/scripts/image_scripts/configure_network_ubuntu.sh
@@ -9,7 +9,7 @@ network:
       dhcp4-overrides:
         use-dns: no
       nameservers:
-        addresses: [1.1.1.1, 1.0.0.1]
+        addresses: [1.1.1.1, 8.8.8.8]
 
 EOF
 # Apply the changes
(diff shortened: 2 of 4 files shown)


ci/scripts/image_scripts/configure_network_centos.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# make NetworkManager to prevent modifying resolv.conf
+sudo sed -i '/^\[main\]/a dns=default' /etc/NetworkManager/NetworkManager.conf
+
+
+
+# Specify DNS servers for systemd-resolved.
+cat <<EOF | sudo tee /etc/systemd/resolved.conf
+[Resolve]
+DNS=8.8.8.8 8.8.4.4
```
tuminoid, line 13: We're using Cloudflare's 1.1.1.1 and 1.0.0.1 as primary DNS on Ubuntu, any reason why we'd default to Google on Centos?

Sunnatillo: There is no specific reason behind. As external dns I specified Google's dns addresses as this is commonly used. For fallback, wanted to use different dns addresses. 
Should we use the same?

tuminoid: I see you split the difference and changed ubuntu and centos to match, and use both. Nice.

tuminoid (approved):

kashifest:
We had similar kind of configuration which we removed here https://github.com/Nordix/metal3-dev-tools/pull/674/ so I am trying to understand why we removed it and what problems would we face if we reintroduce it @Rozzii

Sunnatillo:
> We had similar kind of configuration which we removed here #674 so I am trying to understand why we removed it and what problems would we face if we reintroduce it @Rozzii

We previously disabled whole cloud init. and it Prevented us to set 

> We had similar kind of configuration which we removed here #674 so I am trying to understand why we removed it and what problems would we face if we reintroduce it @Rozzii

We previously disable whole cloud init and it prevented us to use IPAN based IP addresses. @Rozzii removed the changes with his changes. This PR only disabling setting dns address by NetworkManager.

lentzi90:
Wondering the same as @kashifest . We have been going back and forth with this. I don't think writing directly to `resolv.conf` is the right way to go. That will end up with having to make the file immutable to prevent cloud-init from overwriting it and then facing cloud-init errors instead.
https://github.com/Nordix/metal3-dev-tools/pulls?q=is%3Apr+dns+networkmanager+is%3Aclosed

Why don't we go back to something like this instead:
https://github.com/Nordix/metal3-dev-tools/blob/c22b7c85c69cc5d775bcccabd7141749d083bf67/ci/scripts/image_scripts/configure_network_centos.sh

lentzi90:
Another idea is also to configure this in Cleura. We can set the nameservers to use directly on the subnet there.

Rozzii:
We had a discussion and for now this will get a hold as my opinion is that DNS server should be specified via Metal3DataTemplate.

Sunnatillo:
/close

As we discussed it is better to set dns address through metal3datatemplate.

PR #705, opened 2023-10-02
align node image building with packer "retry" workflow

Rozzii:
This commit:
  - Changes "mv" commands to "cp" in order to avoid failures in case of packer "retry" actions
  - Also upgrades the Centos 9 Stream source image as the previous version was a year old.

### DIFF
--- a/ci/images/gen_metal3_centos_image.sh
+++ b/ci/images/gen_metal3_centos_image.sh
@@ -41,7 +41,7 @@ fi
 source "${OS_SCRIPTS_DIR}/utils.sh"
 
 IMAGE_NAME="${CI_IMAGE_NAME}-$(get_random_string 10)"
-SOURCE_IMAGE_NAME="CentOS-Stream-9-20220829"
+SOURCE_IMAGE_NAME="CentOS-Stream-9-20230925"
 USER_DATA_FILE="$(mktemp -d)/userdata"
 SSH_USER_NAME="${CI_SSH_USER_NAME}"
 SSH_KEYPAIR_NAME="${CI_KEYPAIR_NAME}"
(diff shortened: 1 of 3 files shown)


tuminoid:
Still taking too long time.

Do we need to bump Ubuntu as well? Increase timeout until this passes?

Rozzii:
> Still taking too long time.
> 
> Do we need to bump Ubuntu as well? Increase timeout until this passes?

Yeah probably. I have no better idea ATM either.

tuminoid:
> > Still taking too long time.
> > Do we need to bump Ubuntu as well? Increase timeout until this passes?
> 
> Yeah probably. I have no better idea ATM either.

Centos image is only a week old, so timeout seems to be the only option. Unless it is again some stupid DNS slowness causing this. @Sunnatillo might know.

For Ubuntu, we are using a year+ old image. Could bump it to `Ubuntu 22.04 Jammy Jellyfish x86_64` which is only 3 weeks old.

Rozzii:
> > > Still taking too long time.
> > > Do we need to bump Ubuntu as well? Increase timeout until this passes?
> > 
> > 
> > Yeah probably. I have no better idea ATM either.
> 
> Centos image is only a week old, so timeout seems to be the only option. Unless it is again some stupid DNS slowness causing this. @Sunnatillo might know.
> 
> For Ubuntu, we are using a year+ old image. Could bump it to `Ubuntu 22.04 Jammy Jellyfish x86_64` which is only 3 weeks ol

Rozzii:
/cc @lentzi90 @tuminoid @kashifest @smoshiur1237 @mboukhalfa 
Would it be fine to overwrite the test results?

tuminoid (approved):
/lgtm

tuminoid:
> /cc @lentzi90 @tuminoid @kashifest @smoshiur1237 @mboukhalfa Would it be fine to overwrite the test results?

Yes, failure is not due to this PR.

lentzi90 (approved):

PR #716, opened 2023-10-23
Remove unnecessary copy of crictl binary

kashifest:
crictl is already put in place /usr/local/bin [here](https://github.com/Nordix/metal3-dev-tools/blob/main/ci/scripts/image_scripts/install_crio_on_centos.sh#L34) in the install_crio_on_centos.sh script called a line before. Instead of putting it to /usr/bin in the provision_ script, we now put it in /usr/bin right away.


mquhuy:
Hmm. If the binary was already there, doesn't it make sense to just not curl again? (or maybe only curl after verifying that it's not in /usr/local/bin?)

Rozzii (approved):
/approve

smoshiur1237 (approved):

PR #718, opened 2023-10-30
downgrade packer to 1.9.2

Rozzii:
It was observed recently in Metal3 CI, that packer was producing tcp errors at the last stage of the image build process. The images were built and uploaded correctly but packer still finished the build process with a tcp error (sanitized):

```
==> openstack: Error creating image: Post "https://<ostack_api_url>/v3/<proj_id>/volumes/<volume_id>/action":
    read tcp ip:port->ip2:port2: read: connection reset by peer

==> openstack: Provisioning step had errors: Running the cleanup provisioner, if present...

==> openstack: Deleted temporary floating IP '<floating_ip_id>' (<floating_ip>)

==> openstack: Terminating the source server: <vm_id> ...

==> openstack: Error terminating server, may still be around: Resource not found

==> openstack: Waiting for volume packer_<hash> (volume id: <volume_id>) to become available...

==> openstack: Deleting volume: <volume_id> ...

Build 'openstack' errored after 12 minutes 50 seconds: Error creating image:
Post "https://<ostck_url>/v3/<proh_id>/volumes/<volume_id>/action": read tcp ip:port->ip2:port2:
read: connection reset by peer
```

The issue could be caused by a packer bug or incompatibility somewhere between the packer, the packer OpenStack plugin and the OpenStack.

This commit:

- Downgrades the packer version to the previous version in hopes of getting rid of the tcp issue.

### DIFF
--- a/resources/docker/builder/Dockerfile
+++ b/resources/docker/builder/Dockerfile
@@ -1,5 +1,5 @@
 # Get packer executables
-FROM hashicorp/packer:light AS packer
+FROM hashicorp/packer:light-1.9.2 AS packer
 
 # Build openstack client
 FROM python:3.12-alpine@sha256:ae35274f417fc81ba6ee1fc84206e8517f28117566ee6a04a64f004c1409bdac AS builder


Rozzii:
/ test-integration-metal3-dev-tools-ubuntu

kashifest (approved):
/approve

resources/docker/builder/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Get packer executables
-FROM hashicorp/packer:light AS packer
+FROM hashicorp/packer:light-1.9.3 AS packer
```
tuminoid, line 2: Did you check from logs which version of packer we were using? I ask since the tag has been `light` all along, which means it hasn't  been version pinned.

Rozzii: yep, I have checked on docherkhub, what does light mean and it is a secondary tag for light-1.9.4. I have checked that we are building this periodically so we were using the latest reference of "light" so 1.9.4.

tuminoid: 1.9.4 contains single, unrelated bug fix, so it is unlikely to be culprit. However 1.9.3 is doing major changes around the plugins etc (removing all of the rest of bundled plugins), while releasing new `full-1.9.3` image, which all the official plugins pre-installed. I'm thinking we should rather go for `full-1.9.3` to restore the plugins (maybe one of them made a difference), or go back to 1.9.2, which was released less than 4 weeks prior to 1.9.3 and has the old plugins embedded.

https://github.com/hashicorp/packer/blob/main/CHANGELOG.md

Rozzii: Okay , I would say let's go then with `light-1.9.2` as we have never used full and that might introduce other bugs (who knows...).

tuminoid: Sure. Full vs light difference is just official plugins pre-installed. Openstack plugin is by Hashicorp, so it would mean it would come pre-installed and we wouldn't need to run plugin installation later in the Dockerfile.

tuminoid (approved):
/lgtm

PR #719, opened 2023-10-31
disable sshd ClientAlive requests for baseimage

Rozzii:
Currently the Metal3 CI is experiencing a lot of SSH disconnect
issues. This ClientAliveInterval misconfiguration might be a potential
reason for the unstable connections.

This commit:
  - Sets the value to 0 for ClientAliveInterval on
    all images via the base image thus the sshd won't request
    "keep alive" responses. This change is to make sure the
    ssh connection is not closed in any situation.


tuminoid (approved):
Couple notices for future us:
- Setting it to `0` is security-risk in interactive environment, but here it is only automated environment.
- This config hasn't changed since 2019 so it shouldn't change anything
- This is the only place where we set SSH Server config, elsewhere it is all SSH Client configs

/lgtm

kashifest (approved):

smoshiur1237 (approved):
/lgtm

smoshiur1237 (approved):
/lgtm

Rozzii:
CI is not working correctly: 

```
==> openstack: Waiting for SSH to become available...
Cannot contact metal3-worker-317: hudson.remoting.RequestAbortedException: java.io.IOException: Unexpected termination of the channel
Cancelling nested steps due to timeout
```

This change should have no negative effect , and even if it has that would be visible after the image is in wide scale use so I will overwrite the tests.

/override  test-integration-metal3-dev-tools-centos

PR #722, opened 2023-11-03
image building improvements

tuminoid:
1. do not recreate keypair each time, it is conflicting between jobs
2. check if network detection fails, and abort before trying to call packer build, which would explode with cryptic issues"

### DIFF
--- a/ci/images/gen_metal3_ubuntu_image.sh
+++ b/ci/images/gen_metal3_ubuntu_image.sh
@@ -28,6 +28,11 @@ NETWORK="$(get_resource_id_from_name network "${CI_EXT_NET}")"
 FLOATING_IP_NETWORK="$( [ "${USE_FLOATING_IP}" = 1 ] && echo "${EXT_NET}")"
 SSH_USER_GROUP="sudo"
 
+if [[ -z "${NETWORK}" ]]; then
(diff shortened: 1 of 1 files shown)


lentzi90 (approved):
/approve

kashifest (approved):
/approve

tuminoid:
/hold
Let's not merge this yet so it doesn't mess with the tests during the weekend.

tuminoid:
As expected, both builds failed to network errors thanks to CI. This is confirming that these changes do not affect that, but nonetheless these should be fixed. Keeping the hold over weekend.

tuminoid:
Are these turned back on? Let's see...

lentzi90:
Yeah I wonder... should we turn on the PR tests and leave the periodic tests? Then we can try things here when needed without destroying all our nightly jobs :slightly_smiling_face:

tuminoid:
/unhold
If only the CI worked ...

Rozzii (approved):
/lgtm

Rozzii:
I am fine if we override the tests, this PR can be signed off by just human reviewers with high confidence.

tuminoid:
I had high hopes that one by one these would work, like they did in two other PRs, but I guess not.

/override test-integration-metal3-dev-tools-centos

PR #723, opened 2023-11-20
Migrate to k8s community-owned repos.

kashifest:
Migrate to k8s community-owned repo as suggested by https://kubernetes.io/blog/2023/08/15/pkgs-k8s-io-introduction/.

### DIFF
--- a/ci/scripts/image_scripts/provision_node_image_centos.sh
+++ b/ci/scripts/image_scripts/provision_node_image_centos.sh
@@ -4,9 +4,20 @@ set -uex
 
 SCRIPTS_DIR="$(dirname "$(readlink -f "${0}")")"
 export KUBERNETES_VERSION=${KUBERNETES_VERSION:-"v1.28.1"}
+export KUBERNETES_MINOR_VERSION=${KUBERNETES_VERSION%.*}
 export KUBERNETES_BINARIES_VERSION="${KUBERNETES_BINARIES_VERSION:-${KUBERNETES_VERSION}}"
-export KUBERNETES_BINARIES_CONFIG_VERSION=${KUBERNETES_BINARIES_CONFIG_VERSION:-"v0.15.1"}
 export CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
+
+# migrate to the Kubernetes community-owned repositories
+cat <<EOF | sudo tee /etc/yum.repos.d/kubernetes.repo
+[kubernetes]
+name=Kubernetes
+baseurl=https://pkgs.k8s.io/core:/stable:/${KUBERNETES_MINOR_VERSION}/rpm/
+enabled=1
+gpgcheck=1
+gpgkey=https://pkgs.k8s.io/core:/stable:/${KUBERNETES_MINOR_VERSION}/rpm/repodata/repomd.xml.key
+exclude=kubelet kubeadm kubectl cri-tools kubernetes-cni
+EOF
 
 "${SCRIPTS_DIR}"/install_crio_on_centos.sh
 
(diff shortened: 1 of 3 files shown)


ci/scripts/image_scripts/provision_node_image_centos.sh
```diff
@@ -4,10 +4,22 @@ set -uex
 
 SCRIPTS_DIR="$(dirname "$(readlink -f "${0}")")"
 export KUBERNETES_VERSION=${KUBERNETES_VERSION:-"v1.28.1"}
+export KUBERNETES_MINOR_VERSION=${KUBERNETES_VERSION%.*}
 export KUBERNETES_BINARIES_VERSION="${KUBERNETES_BINARIES_VERSION:-${KUBERNETES_VERSION}}"
 export KUBERNETES_BINARIES_CONFIG_VERSION=${KUBERNETES_BINARIES_CONFIG_VERSION:-"v0.15.1"}
 export CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
 
+# migrate to the Kubernetes community-owned repositories
+cat <<EOF | sudo tee /etc/yum.repos.d/kubernetes.repo
```
kashifest, line 13: I was a bit unsure if changing yum repos would take any effect for dnf , but if I understand this correctly here https://dnf.readthedocs.io/en/latest/conf_ref.html#:~:text=%60DNF%60_%20by%20default%20uses,repos. it does.

ci/scripts/image_scripts/provision_node_image_ubuntu.sh
```diff
@@ -47,20 +47,18 @@ sudo chmod +x /usr/local/bin/retrieve.configuration.files.sh
...
 "${SCRIPTS_DIR}"/install_crio_on_ubuntu.sh
 
 echo  "Installing kubernetes binaries"
-curl -L --remote-name-all "https://storage.googleapis.com/kubernetes-release/release/${KUBERNETES_BINARIES_VERSION}/bin/linux/amd64/{kubeadm,kubelet,kubectl}"
-sudo chmod a+x kubeadm kubelet kubectl
-sudo cp kubeadm kubelet kubectl /usr/local/bin/
-sudo mkdir -p /etc/systemd/system/kubelet.service.d
-sudo retrieve.configuration.files.sh https://raw.githubusercontent.com/kubernetes/release/"${KUBERNETES_BINARIES_CONFIG_VERSION}"/cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service /etc/systemd/system/kubelet.service
-sudo retrieve.configuration.files.sh https://raw.githubusercontent.com/kubernetes/release/"${KUBERNETES_BINARIES_CONFIG_VERSION}"/cmd/kubepkg/templates/latest/deb/kubeadm/10-kubeadm.conf /etc/systemd/system/kubelet.service.d/10-kubeadm.conf
+sudo apt-get install -y kubelet kubeadm kubectl
```
kashifest, line 60: I couldn't find a way to specify a patch version here, so I kept it as it is. The downside of this is it will always download the latest patch version.  The other workaround is to not install it through OS packages but download it.

tuminoid: I don't like installing any version, even if its a patch version. They are known to have issues in the patches sometimes.

The package version format seems to be `v1.28.1-1.1` (ie. `-1.1` suffix), but if we don't trust that, then we can use `apt list` for example:
`apt list -a kubectl 2>/dev/null | grep -E "(^kubectl/unknown ${KUBERNETES_VERSION/v/}.*amd64$)" | head -1 | cut -f2 -d" "`
(note that is must use `apt`, not `apt-get` since apt-get does not have `list`)

and then we can install it `apt-get install -y kubelet=1.28.1-1.1`

kashifest: I tried with the suffix, but it seems its not the same for different patch versions, I can try with the apt list, if we dont want to download it.

tuminoid (commented):
LGTM, let's just get the Centos test to pass so we know it works on that side as well.

ci/scripts/image_scripts/provision_node_image_ubuntu.sh
kashifest: As discussed offline , instead of apt-list , we now stick to -* for the build version

Sunnatillo (approved):
/lgtm

lentzi90 (approved):
/approve

PR #724, opened 2023-11-27
Container building pipeline: supports git tags and remove local image info database

mquhuy:
With this PR, container building script will handle git tags, provided as `refs/tags/<tag_name>` by github actions. The exact container tags are as followed:

- If checking a git tag, the only image tag generated will be the tag name
- If checking a git branch, the following image tags are appointed to the same newly-built image:
   - The branch name. Eg. `release-0.4`
   - The branch name with date and latest short hash: E.g. `release-0.4_20231127_d03da556`
   - If the branch is `main`, tagging it as `latest`.

Edit: Changed some implementation to take into account new parameters set implemented in https://gerrit.nordix.org/c/infra/cicd/+/20053/3/jjb/metal3/job_container_image_building.yaml. Basically this means no image information will be stored in this repo, but got piped from upstream (jjb configuration)


scripts/build-container-image.sh
```diff
@@ -47,22 +65,17 @@ git_get_current_commit_short_hash() {
...
 get_date() {
   # Should this be current date, or latest commit date?
   date +%Y%m%d
 }
 
-get_image_tag() {
-  echo "$(git_get_current_branch)_$(get_date)_$(git_get_current_commit_short_hash)"
-}
-
 cleanup_old_tags() {
```
tuminoid, line 73:
```suggestion
cleanup_old_tags() {
  declare -a branch_tags=()
  declare -a all_tags=()
  declare -a sorted_existing_tags=()
```
Start by declaring your arrays.

scripts/build-container-image.sh
```diff
@@ -47,22 +65,17 @@ git_get_current_commit_short_hash() {
...
-  all_tags=$(list_tags)
-  branch=$(git_get_current_branch)
+  if [[ "${GIT_REFERENCE_TYPE}" != "branch" ]]; then
+    return
+  fi
+  all_tags=($(list_tags))
```
tuminoid, line 77:
```suggestion
  mapfile -t all_tags < <(list_tags)
```
Assigning the command output to an array is problematic, shellcheck recommends `mapfile` which works here nicely.

https://www.shellcheck.net/wiki/SC2207

scripts/build-container-image.sh
```diff
@@ -47,22 +65,17 @@ git_get_current_commit_short_hash() {
...
+  branch="${GIT_REFERENCE}"
 
   declare -a branch_tags=()
```
tuminoid, line 81: L87 (GH doesn't allow me to comment on that, since it is not changed) is buggy as well.

```suggestion
  # shellcheck: disable=SC2207
  sorted_existing_tags=($(sort <<<"${branch_tags[*]}"))
```
Ie. remove set of `"`, which would make the sort output a single string, and remove the `IFS=$'\n'` since we're not line-feed delimited, but whitespace delimited.

scripts/build-container-image.sh
```diff
@@ -81,24 +94,30 @@ cleanup_old_tags() {
...
   image_tag=${1:?}
   echo "${HUB}/${HUB_ORG}/${IMAGE_NAME}:${image_tag}"
 }
 
 build_container_image() {
-  image_tag=$(get_image_tag)
-  image_latest_tag=$(get_image_latest_tag)
-  image_path=$(get_image_path "${image_tag}")
-  image_latest_path=$(get_image_path "${image_latest_tag}")
+  image_path=$(get_image_path "${GIT_REFERENCE}")
```
tuminoid, line 103:
```suggestion
  declare -a image_tags=()

  image_path=$(get_image_path "${GIT_REFERENCE}")
```
You want to declare your arrays.

tuminoid (commented):
The array handling is broken + some nits.

scripts/build-container-image.sh
tuminoid: Sorry, the disable line should read `# shellcheck disable=SC2207` without the `:`

scripts/build-container-image.sh
```diff
@@ -29,40 +46,32 @@ list_tags() {
...
+    return
+  fi
+  declare -a branch_tags=()
+  declare -a all_tags=()
+  declare -a sorted_existing_tags=()
+
+  mapfile -t all_tags < <(list_tags)
+  branch="${GIT_REFERENCE}"
 
   declare -a branch_tags=()
```
tuminoid, line 76: This is now duplicate.

scripts/build-container-image.sh
```diff
@@ -114,7 +130,7 @@ build_image() {
   git clone "${repo_link}" "${repo_name}"
   dockerfile_directory=$(jq < "${IMAGES_JSON}" -r .\""${IMAGE_NAME}"\".dockerfile_location)
   cd "${repo_name}"
-  git checkout "${BRANCH_NAME}"
+  git checkout "${GIT_REFERENCE}"
   cd "${REPO_LOCATION}/${repo_name}${dockerfile_directory}"
   build_container_image
 }
```
tuminoid, line 136: L140, comment out the cleanup part and add a TODO comment, since we're going to finish that in separate PR.

scripts/build-container-image.sh
```diff
@@ -12,13 +12,30 @@ set -o pipefail
...
 __dir__=$(realpath "$(dirname "$0")")
 IMAGES_JSON="${__dir__}/files/container_image_names.json"
 
+case "${GIT_REFERENCE}" in
+  *"refs/heads"*)
+    GIT_REFERENCE_TYPE="branch"
+    GIT_REFERENCE=$(echo "${GIT_REFERENCE}" | sed 's/refs\/heads\///')
+    echo "Detected branch name: ${GIT_REFERENCE}"
+    ;;
+  *"refs/tags"*)
```
tuminoid, line 28: I just remembered that we do not want to build images for all tags. In Quay, we build only tags in format `v1.x.x`, so `apis/1.x.x` etc get ignored. So either ignore by tag name (needs to start with `^v`, otherwise ignore), or alternatively build only annotated tags, and not lightweight tags.

mquhuy: imo those kinds of logics should be put into the github actions, which can/will be updated more frequently and more easily.

mquhuy: And I actually found several tags in the form `apis_1.x.x` in quay? Remember vaguely that Adil asked for that at some point.

mquhuy: Well, you can configure it by setting "ignore" tags: https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-filters-to-target-specific-branches-or-tags-for-push-events

Other than our convenience in , the difference is that: if you config it in github actions, the tags are actually ignored, but if you config it in this script, the jobs are still triggered on jenkins (hence jenkins worker hardware still consumed), just that there would be no image built.

About convenience, we're sure that we only want `v...` tags to be built in ALL containers, then updating it here is more convenient, otherwise it would be a bit hacky imo. All in all, I'd still prefer to have these ignored patterns specified in the github actions, but if you still insist to put it here, it's fine by me.

tuminoid: Nah. Since we can configure them properly, and it saves a job, then  let's put them in the action then. Like said, it then allows different tag patterns per repo, even if we currently have them exactly the same for all.

kashifest (dismissed):

smoshiur1237 (commented):
/lgtm

mquhuy:
> I mentioned in one of the GH workflow file PRs that we could pass the information about Dockerfile and image name via the env, instead of maintaining the JSON file in this repo.
> 
> Changing it later requires a lot more effort and syncing PRs than fixing it right now. WDYT? @mquhuy @kashifest @lentzi90

Just for reference, here is the link to @tuminoid  's original comment: https://github.com/metal3-io/baremetal-operator/pull/1456#discussion_r1412022826

I have put in there some of my ideas, in this thread I can also mention that changing the structure now is not easy, either:
- We need to add many more params to the jenkins jobs.
- The jenkins jobs are currently designed to be separated (i.e. each image has one job pipeline). Having them combined is possible, but takes time, and I still think having them separated is better for management.

tuminoid:
> > I mentioned in one of the GH workflow file PRs that we could pass the information about Dockerfile and image name via the env, instead of maintaining the JSON file in this repo.
> > Changing it later requires a lot more effort and syncing PRs than fixing it right now. WDYT? @mquhuy @kashifest @lentzi90
> 
> Just for reference, here is the link to @tuminoid 's original comment: [metal3-io/baremetal-operator#1456 (comment)](https://github.com/metal3-io/baremetal-operator/pull/1456#discussion_r1412022826)
> 
> I have put in there some of my ideas, in this thread I can also mention that changing the structure now is not easy, either:
> 
> * We need to add many more params to the jenkins jobs.

The JSON file has 3 parameters: repository, dockerfile location, and repo_name (image name). They are pretty specific per workflow job. You're already passing one of them (image name), so 2 new.

> * The jenkins jobs are currently designed to be separated (i.e. each image has one job pipeline). Having them combined is possible, but takes time, and I still think having them separated is better for management.

I'm not following this argument. I'm suggesting adding 2 more environment variables to the workflow files, and removing the `build_image()` function (which basically reads the same information from the file as we provided in the env). How would this affect the jenkins jobs in any way?

mquhuy:
> > * The jenkins jobs are currently designed to be separated (i.e. each image has one job pipeline). Having them combined is possible, but takes time, and I still think having them separated is better for management.
> 
> I'm not following this argument. I'm suggesting adding 2 more environment variables to the workflow files, and removing the `build_image()` function (which basically reads the same information from the file as we provided in the env). How would this affect the jenkins jobs in any way?

I'm answering your later point first, since it's needed to understand the first point. If you check our jenkins, you will find several jobs in form `metal3_<image_name>_container_image_build`. Each of them represents a separate container image build.
Now if we wanted to remove the centralized information, we would no longer be able to manage them separately. Each image build would represent one run of the same jenkins job, and we would only know which image got built by checking the run params. 

> > 
> > * We need to add many more params to the jenkins jobs.
> 
> The JSON file has 3 parameters: repository, dockerfile location, and repo_name (image name). They are pretty specific per workflow job. You're already passing one of them (image name), so 2 new, and for 95% of the cases, we can default the Dockerfile to be `/Dockerfile`. Effectively 1 env.

I don't think it's that simple, we should not just care about the number of params, but also their complexity
- The `image name` is currently there, but it's the same with the job name (and the job name is the default), so it's hardly mistaken. The only thing you need to care is the branch/tag name.
- In a new setup when you use the same job for all builds, even the image name can be a tricky thing to put in, not to mention stuff like url. And remember that the script has no way to check if an image name is legit, it will build and push anything as long as the repo can be cloned and it has a Dockerfile.

lentzi90:
I agree with @tuminoid . Getting rid of the json file and using parameters instead makes the logic simpler and the job more generic/re-usable.

> In a new setup when you use the same job for all builds, even the image name can be a tricky thing to put in, not to mention stuff like url. And remember that the script has no way to check if an image name is legit, it will build and push anything as long as the repo can be cloned and it has a Dockerfile.

I don't think it has to be all the same job, but even if we do it that way, can you explain why this would be problematic @huy? We pass URLs and names, tags, branches, etc as parameters all the time.

mquhuy:
> I agree with @tuminoid . Getting rid of the json file and using parameters instead makes the logic simpler and the job more generic/re-usable.
> 
> > In a new setup when you use the same job for all builds, even the image name can be a tricky thing to put in, not to mention stuff like url. And remember that the script has no way to check if an image name is legit, it will build and push anything as long as the repo can be cloned and it has a Dockerfile.
> 
> I don't think it has to be all the same job, but even if we do it that way, can you explain why this would be problematic @huy? We pass URLs and names, tags, branches, etc as parameters all the time.

I'm not saying that it'd be unbearable, but I don't see why we should do that? Yes, I agree that the job would be more reusable, because now we can just put any repo url and it will build it. But do we really want to allow the job to accept any repo and build anything, and push it all to our quay?
From my understanding, if you want to reuse the job for different kinds of images, it should be in the same job, simply because we can't separate them without knowing how many and what they are in the first place. And that's also my main motivation for the setup we're having, I think the json logic keeps it more manageable: we have separate jobs (on jenkins) to keep track of the builds of different images, and we have a list of images that the script can build. (And for manual builds, we don't have to put in multiple more params)

tuminoid:
As discussed off-line, JSON is to be removed, as it contains no information that cannot be in the workflow file, or the pipeline file, and hence is just another file that has to be kept synced. It also makes sense to do this change before merging the script or other PRs to avoid doing double work.

Rozzii (approved):
/lgtm

scripts/build-container-image.sh
```diff
@@ -29,96 +46,53 @@ list_tags() {
...
-  image_latest_tag=$(get_image_latest_tag)
+  image_tag=$(echo "${GIT_REFERENCE}" | sed 's/\//_/')
   image_path=$(get_image_path "${image_tag}")
-  image_latest_path=$(get_image_path "${image_latest_tag}")
+  echo "Building the image as ${image_path}"
   docker build -t "${image_path}" .
-  docker push "${image_path}"
-  docker tag "${image_path}" "${image_latest_path}"
-  docker push "${image_latest_path}"
+  # docker push "${image_path}"
```
tuminoid, line 68: Are the docker pushes intentionally commented out?
/hold

mquhuy: Ah, no. I must have missed it, initially commented it just to test the script. Thanks for the notice :D

Rozzii:
/ test-integration-metal3-dev-tools-centos
/test-integration-metal3-dev-tools-ubuntu

tuminoid (approved):
/lgtm
/unhold

tuminoid:
> /test-integration-metal3-dev-tools-centos /test-integration-metal3-dev-tools-ubuntu

/override test-integration-metal3-dev-tools-centos

kashifest (approved):
/approve

PR #729, opened 2023-12-04
Cleanup unused linter test related files

Sunnatillo:
Removes:
 - linter test pipeline
 - lint related docker image building scripts


tuminoid:
Can you fix the Git commit message? Now it says `Remove outdated:`  which is not right.

lentzi90 (approved):
/approve

Sunnatillo:
> Can you fix the Git commit message? Now it says `Remove outdated:` which is not right.

done

tuminoid:
> > Can you fix the Git commit message? Now it says `Remove outdated:` which is not right.
> 
> done

Now it says: `cleanups unused:` so not better. You need to add title line, the current commit message goes into the body.

Sunnatillo:
> > > Can you fix the Git commit message? Now it says `Remove outdated:` which is not right.
> > 
> > 
> > done
> 
> Now it says: `cleanups unused:` so not better. You need to add title line, the current commit message goes into the body.

it says: 
cleanups unused:
 - lint pipeline
 - go lint related docker image build scripts

lentzi90:
> it says: cleanups unused:
> 
>     * lint pipeline
> 
>     * go lint related docker image build scripts

The problem is that this is not just one big text field. The first line is like a title and the title `cleanups unused:` does not make sense.

image

Sunnatillo:
> > it says: cleanups unused:
> > ```
> > * lint pipeline
> > 
> > * go lint related docker image build scripts
> > ```
> 
> The problem is that this is not just one big text field. The first line is like a title and the title `cleanups unused:` does not make sense.
> 
> image
> 
> See https://git-scm.com/docs/git-commit#_discussion

I see. I have changed it now. Thanks for clarifying.

tuminoid:
> > > it says: cleanups unused:
> > > ```
> > > * lint pipeline
> > > 
> > > * go lint related docker image build scripts
> > > ```
> > 
> > 
> > The problem is that this is not just one big text field. The first line is like a title and the title `cleanups unused:` does not make sense.
> > image
> > See https://git-scm.com/docs/git-commit#_discussion
> 
> I see. I have changed it now. Thanks for clarifying.

No need to change it further, but Git best practices say the title line should be 50 characters or less, and the body should be 72 characters per line. Just FYI.

tuminoid (commented):
/lgtm

tuminoid (approved):

PR #737, opened 2024-01-09
Test packer disconnect config

Sunnatillo:


Sunnatillo:
//test-integration-metal3-dev-tools-ubuntu

tuminoid:
The network issues we're having are not packer issues nor solvable in packer. While we may be able to fix a build sometimes, I don't think this is worth the effort.

Sunnatillo:
> The network issues we're having are not packer issues nor solvable in packer. While we may be able to fix a build sometimes, I don't think this is worth the effort.

Agree.

PR #755, opened 2024-03-22
Fix node image building

Sunnatillo:


Sunnatillo:
/close
We have new workflow in project infra

PR #766, opened 2024-07-02
Move Artifact cleanup and Nordix repos sync jobs to dynamic workers

Sunnatillo:
This PR moves artifact cleanup and Nordix repos sync job to dynamic workers.

### DIFF
--- a/ci/jobs/artifact_cleanup.pipeline
+++ b/ci/jobs/artifact_cleanup.pipeline
@@ -13,7 +13,7 @@ script {
 }
 
 pipeline {
-  agent { label 'metal3-static-workers' }
+  agent { label 'metal3-medium-ubuntu' }
   environment {
     METAL3_CI_USER="metal3ci"
     VM_KEY = "${VM_KEY}"
--- a/ci/jobs/update_nordix_repos.pipeline
+++ b/ci/jobs/update_nordix_repos.pipeline
@@ -3,7 +3,7 @@ ci_git_branch = "main"
 ci_git_credential_id = "nordix-metal3-github-ssh"
 
 pipeline {
-  agent { label 'metal3-static-workers' }
+  agent { label 'metal3ci-4c16gb-ubuntu' }
   environment {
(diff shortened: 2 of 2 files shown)


ci/jobs/update_nordix_repos.pipeline
```diff
@@ -3,7 +3,7 @@ ci_git_branch = "main"
 ci_git_credential_id = "nordix-metal3-github-ssh"
 
 pipeline {
-  agent { label 'metal3-static-workers' }
+  agent { label 'metal3ci-4c16gb-ubuntu' }
```
tuminoid, line 6: Is this the smallest one? The job is just pulling and pushing some git repos, so it can be basically as small as we have...

ci/jobs/artifact_cleanup.pipeline
```diff
@@ -13,7 +13,7 @@ script {
 }
 
 pipeline {
-  agent { label 'metal3-static-workers' }
+  agent { label 'metal3-medium-ubuntu' }
```
tuminoid, line 16: Same here I think.

ci/jobs/update_nordix_repos.pipeline
Sunnatillo: yes, unfortunately this is the smallest one for now.

ci/jobs/artifact_cleanup.pipeline
Sunnatillo: same. We do not have small vms.
The idea is to use same labels for most of the tests. This will reduce the number of dynamic workers type that needs to be maintained.

tuminoid: OK, we can live with that.

ci/jobs/update_nordix_repos.pipeline
tuminoid: OK, we'll live with that.

tuminoid (commented):
/approve

PR #767, opened 2024-07-03
Replace the old ssh key with a new one

kashifest:
This is part of the CI key rotation. We are replacing the old ssh key with the new one.
Signed-off-by: Kashif Khan <kashif.khan@est.tech>


tuminoid (commented):
/approve

We should stop injecting keys in like this.

Rozzii (commented):
/lgtm

PR #770, opened 2024-08-28
:seedling: Remove release 1.5

mquhuy:
/hold

### DIFF
--- a/scripts/update-nordix-repos-master.sh
+++ b/scripts/update-nordix-repos-master.sh
@@ -153,16 +153,16 @@ for repo in ${UPDATE_REPO}; do
 done
 
 # Sync non default branches on selected repos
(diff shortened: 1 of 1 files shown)


tuminoid:
This again is doing two things at the same time. For sync job this is fine, but make the title and description about switching the release, and not just mention removing.

tuminoid:
> This again is doing two things at the same time. For sync job this is fine, but make the title and description about switching the release, and not just mention removing.

OK, this is two commit PR again, which one is already in another PR to be first merged. So we're rebasing and then removing.

tuminoid (commented):
/approve
Let's keep it held until we've cut final releases from 1.5 branches

kashifest (commented):
/lgtm

Sunnatillo (approved):
/approve

PR #781, opened 2025-03-19
Remove @mquhuy from approvers

kashifest:
@mquhuy is not active as an approver in Nordix repos anymore. Thanks Huy for all your contributions.
Signed-off-by: Kashif Khan <kashif.khan@est.tech>


kashifest:
I have left the renovate.json file untouched since that needs to be taken up separately if we need it here or not.

tuminoid (commented):
/approve

Rozzii (commented):
/lgtm

PR #789, opened 2025-05-28
stop syncing some old repos

tuminoid:
Stop syncing nmstate, topolvm and some sriov repos, as they're not often used and don't need syncing every 30 mins.  Also stop syncing locked release-1.8 branches.

Cloning and syncing just creates unnecessary load on Github and CI.


tuminoid:
/cc @kashifest @lentzi90 @adilGhaffarDev @Rozzii 
WDYT? I don't see much use syncing these repos, but I can adjust if you feel differently.

kashifest:
> /cc @kashifest @lentzi90 @adilGhaffarDev @Rozzii WDYT? I don't see much use syncing these repos, but I can adjust if you feel differently.

Hmmm, AFAIR and you folks can correct me if I am wrong, these repos are  still used in downstream, so at some point there might come something to fix or patch, you never know. I am of the opinion of keeping the syncs for the repos although I agree syncing every 30 minutes might be too much. Anyways, I feel this is not a heavy job anyhow.

Rozzii:
I don't see any benefit by removing the sync, I would keep it, who knows we might need to work on these again.

tuminoid:
Well the job runs every 30 mins, there is no adjusting if we want to have it like that for the most repos.

/close

PR #796, opened 2025-08-29
Small adjustments to jump host scripts

terror96:
A few jump host specific scripts do not bail out in case of an error, because they pipe commands. Added a flag to enforce error with piped commands.

Also, introduced a new variable to specify a potentially unique image for the jump host.


ci/scripts/openstack/infra_defines.sh
```diff
@@ -34,3 +34,4 @@ DEV_INT_SUBNET_CIDR="10.1.10.0/24"
 DEV_JUMPHOST_NAME="metal3-dev-jumphost"
 DEV_JUMPHOST_FLOATING_IP_TAG="metal3_dev_jumphost_fip"
 DEV_JUMPHOST_EXT_SG="metal3_dev_jumphost_ext_sg"
+DEV_JUMPHOST_IMAGE="metal3-ci-ubuntu-jenkins-img"
```
tuminoid, line 37: I guess we want to change to hardened jumphost image or vanilla image and not the CI image as it contains way too much SW , but we can do that in another PR.

